PR #34, opened 2023-10-23
Update gRPC for C#

### LINKED ISSUE #9: Switch to a maintained gRPC library for C#
In the C# library, we depend on the gRPC for C# library that [is deprecated](https://grpc.io/blog/grpc-csharp-future/) and [does not officially support Unity](https://github.com/grpc/grpc/issues/22485#issuecomment-1116206421). In practice,  we cannot build Narupa for Unity on arm64, which we need for Meta Quest support.

The recommended library, nowadays, is [gRPC for dotnet](https://github.com/grpc/grpc-dotnet). Therefore, we need to translate our C# codebase to that library.

jbarnoud:
Fixes #9

This may be a breaking change!

* [ ] Validate the change in the NarupaUnityPlugin

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,13 +92,30 @@ jobs:
         run: black --diff --check ./python-libraries
   csharp-tests:
     name: C# unit and integration tests
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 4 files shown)


Ragzouken:
Replaced by https://github.com/IRL2/nanover-protocol/pull/47

PR #49, opened 2024-03-27
switch to netifaces2

Ragzouken:


jbarnoud:
Some of the errors, sadly not all of them, should disappear after a rebase (or a merge) against the main branch.

Ragzouken:
switching to https://pypi.org/project/ifaddr/ may be preferable but needs some investigation to figure out getting equivalent information from it

Ragzouken:
netifaces2 doesn't provide broadcast address, which we need https://github.com/SamuelYvon/netifaces-2/blob/dev/src/win.rs#L31

Issue #54, opened 2024-04-02
Add constant interaction force type


Ragzouken:
Seems to have be useful to have an interaction force that's constant with respect to interaction distance, depending only on explicit force strength and direction to interaction point.

PR #58, opened 2024-04-02
"constant" interaction force

Ragzouken:
closes https://github.com/IRL2/nanover-protocol/issues/54

### DIFF
--- a/python-libraries/nanover-core/src/nanover/imd/imd_force.py
+++ b/python-libraries/nanover-core/src/nanover/imd/imd_force.py
@@ -263,6 +263,7 @@ def calculate_constant_force(
 ) -> Tuple[float, npt.NDArray]:
     """
     Applies a constant force that is independent of the distance between the particle and the interaction site.
+    
     :param particle_position: The position of the particle.
     :param interaction_position: The position of the interaction.
     :param periodic_box_lengths: Vector of periodic boundary lengths.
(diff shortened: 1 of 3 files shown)


python-libraries/nanover-core/src/nanover/imd/imd_force.py
```diff
@@ -8,13 +8,23 @@
...
 
 
+class ForceCalculator(Protocol):
+    def __call__(
+        self,
+        positions: npt.NDArray,
+        interaction_position: npt.NDArray,
+        *,
+        periodic_box_lengths: Optional[npt.NDArray],
+    ) -> Tuple[float, npt.NDArray]: ...
```
Ragzouken, line 25: Maybe I have to change this to `**kwargs: Unpack[...]` or something to get what we really want, but either way this whole thing is exposing that some of the force calculators have parameters that are never varied in the server and aren't covered by tests as far as I can see (k and sigma).

jbarnoud: I think we can put `k` and `sigma` in the function body. We do not plan on exposing these variables.

python-libraries/nanover-core/src/nanover/imd/imd_force.py
```diff
@@ -244,6 +256,26 @@ def calculate_spring_force(
     return energy, force
 
 
+def calculate_constant_force(
+    particle_position: npt.NDArray,
+    interaction_position: npt.NDArray,
+    periodic_box_lengths: Optional[npt.NDArray] = None,
+) -> Tuple[float, npt.NDArray]:
+    """
+    Applies a constant force that is independent of the distance between the particle and the interaction site.
```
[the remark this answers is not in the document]
jbarnoud: Missing the empty line would break the doc rendering.

jbarnoud (approved):
Just a tiny issue with doc rendering but otherwise, LGTM.

PR #59, opened 2024-04-02
Use standard scale for spring constant k in spring force

Ragzouken:
For whatever reason our spring interaction equations were out by a factor of two, meaning that our spring constant k is actually corresponds to 2x the spring constant in the standard form (http://hyperphysics.phy-astr.gsu.edu/hbase/pespr.html). Adjusting the default constant to 2 retains the same energy/force when k is not specified (which is all cases, as far as I can tell).


jbarnoud (approved):
Good to go once the conflicts with #58 are resolved.

PR #62, opened 2024-04-03
 Increase minimum python version to 3.11

Ragzouken:


Ragzouken:
When I run the tests locally, this one always fails:

```
________________________________________ test_subscribe_unused_channel_connect ________________________________________

channel = <grpc._channel.Channel object at 0x0000021CA43A68D0>
connectivity_recorder = ConnectivityRecorder(history=[<ChannelConnectivity.IDLE: (0, 'idle')>, <ChannelConnectivity.CONNECTING: (1, 'connectin..., 'transient failure')>], callback=<function connectivity_recorder.<locals>.record_connectivity at 0x0000021CA4701E40>)

    def test_subscribe_unused_channel_connect(channel, connectivity_recorder):
        """
        Test that subscribing a unused channel's connectivity and forcing connection
        when no corresponding server exists calls the callback first with `IDLE` and
        ends with a `TRANSIENT_FAILURE` state, potentially entering `CONNECTING`
        state between them.
        """
        subscribe_channel_connectivity_change(
            channel, connectivity_recorder.callback, force_connection=True
        )
        time.sleep(NOMINAL_WAIT_TIME)
        assert connectivity_recorder.history[0] == ChannelConnectivity.IDLE
>       assert connectivity_recorder.history[-1] == ChannelConnectivity.TRANSIENT_FAILURE
E       AssertionError: assert <ChannelConnectivity.CONNECTING: (1, 'connecting')> == <ChannelConnectivity.TRANSIENT_FAILURE: (3, 'transient failure')>
E        +  where <ChannelConnectivity.TRANSIENT_FAILURE: (3, 'transient failure')> = ChannelConnectivity.TRANSIENT_FAILURE

python-libraries\nanover-core\tests\utilities\test_grpc_utilities.py:66: AssertionError
```

Ragzouken:
I increased the nominal wait time for the test. It makes sense as a solution because 3.11 is supposed to improve `time.sleep` on windows and I guess on earlier python versions on my system it would have been oversleeping.

README.md
```diff
@@ -80,7 +80,7 @@ The `csharp-libraries/NanoVer.Protocol` folder contains C# implementations of cl
 
 * Install Anaconda (avoid Anaconda 2.7 as it is outdated)
```
Ragzouken, line 81: This warning about 2.7 is probably irrelevant by now?

jbarnoud: Indeed, the 2.7 version does not seem to be on the as prominent anymore.

jbarnoud:
@rhoslynroebuck Could you look at what version of python you are using for the SubtleGame? I guess you are already using 3.11 or higher. I just want to make sure we do not break your use case by accident.

To get the python version, run `python --version` in your environment.

rhoslynroebuck:
I'm using Python 3.11.8

jbarnoud:
Awesome! Ready to go, then!

PR #68, opened 2024-04-04
Client app resets the frame when frame index is 0

### LINKED ISSUE #50: Reset aggregated frame when frame index is 0
Both the rust server and the NanoverUnityPlugin now interpret a frame index of 0 as a signal that the aggregated frame needs to be reset. The python client should follow the same convention to 1) be in line with the rest of the ecosystem and 2) avoid edge cases when the simulation changes.

Relevant change in the unity plugin: https://github.com/IRL2/NanoverUnityPlugin/issues/13
Relevant code in the rust server: https://github.com/IRL2/nanover-rs/blob/1020c28485e6f9f73464bd8d2bea8673ca832c0a/nanover-proto/src/frame.rs#L86-L92

The change should probably occur here: https://github.com/IRL2/nanover-protocol/blob/bf1abbc9c7a645ad1be8b7308e8109bb4aaf7914/python-libraries/nanover-core/src/nanover/app/client.py#L885-L889

A server to demonstrate the issue when the simulation changes:

```python
#!/usr/bin/env python

# coding: utf-8

from nanover.app import NanoverFrameApplication
from nanover.mdanalysis.converter import mdanalysis_to_frame_data
import MDAnalysis as mda
import time
import itertools

u1 = mda.fetch_mmtf('1bta')
u2 = mda.fetch_mmtf('3f1v')

fd1 = mdanalysis_to_frame_data(u1, topology=True, positions=False)
fd2 = mdanalysis_to_frame_data(u1, topology=False, positions=True)
fd3 = mdanalysis_to_frame_data(u2, topology=True, positions=False)
fd4 = mdanalysis_to_frame_data(u2, topology=False, positions=True)

with NanoverFrameApplication.basic_server() as app:
    for index, frame in itertools.cycle(((0, fd1), (1, fd2), (0, fd3), (1, fd4))):
        app.frame_publisher.send_frame(index, frame)
        time.sleep(0.5)
```

jbarnoud:
Fixes #50

### DIFF
--- a/python-libraries/nanover-core/src/nanover/app/client.py
+++ b/python-libraries/nanover-core/src/nanover/app/client.py
@@ -885,9 +885,11 @@ class NanoverImdClient:
         self._are_framed_subscribed = True
 
     def _on_frame_received(self, frame_index: int, frame: FrameData):
-        if self._first_frame is None:
+        if self._first_frame is None or frame_index == 0:
             self._first_frame = frame
         self._frames.append(frame)
+        if frame_index == 0:
+            self._current_frame = FrameData()
         self._current_frame.raw.MergeFrom(frame.raw)
 
     def __enter__(self):
--- a/python-libraries/nanover-core/tests/test_client.py
+++ b/python-libraries/nanover-core/tests/test_client.py
@@ -205,9 +205,14 @@ def test_reconnect_receive(client_server, simple_frame_data):
     assert client.latest_frame is not None
 
 
-def test_frame_reset(client_server, simple_frame_data, disjoint_frame_data):
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


Ragzouken:
Looks good. May be worth a test that uses `subscribe_to_frames` as well as `subscribe_to_all_frames`, I don't expect any difference for the client since they use the same frame merging method, but I could imagining it catching problems where the server has a bug in aggregating frames or something.

jbarnoud:
> Looks good. May be worth a test that uses `subscribe_to_frames` as well as `subscribe_to_all_frames`, I don't expect any difference for the client since they use the same frame merging method, but I could imagining it catching problems where the server has a bug in aggregating frames or something.

Done :)

Issue #70, opened 2024-04-08
Cannot access "current_frame" from a python client


rhoslynroebuck:
Receive the below error when attempting to access `current_frame` from a nanover client. Note that `_current_frame` is accessible and populated as expected. 

![image](https://github.com/IRL2/nanover-protocol/assets/50617818/283b59f7-9b1c-4dcc-bfe7-1c23128b3d7f)

Ragzouken:
More info: this probably only shows up with the rust server, which has a `'forces.user.index': []` field in the frame. The frame cloning code doesn't support empty arrays because it doesn't know what array type they should be. Not exactly sure why at this point.

rhoslynroebuck:
Still getting the following error when accessing `client.current_frame`:

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
File ~\anaconda3\envs\subtle-game\lib\site-packages\nanover\trajectory\frame_data.py:508, in ArraysView.set(self, key, value)
    507 try:
--> 508     reference_value = value[0]
    509 except IndexError:

File ~\anaconda3\envs\subtle-game\lib\site-packages\google\protobuf\internal\containers.py:93, in BaseContainer.__getitem__(self, key)
     92 """Retrieves item by the specified key."""
---> 93 return self._values[key]

IndexError: list index out of range

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 client.current_frame

File ~\anaconda3\envs\subtle-game\lib\site-packages\nanover\app\client.py:61, in _need_attribute.<locals>.wrapper(self, *args, **kwargs)
     59 if getattr(self, attr) is None:
     60     raise RuntimeError(f"Not connected to {name} service")
---> 61 return func(self, *args, **kwargs)

File ~\anaconda3\envs\subtle-game\lib\site-packages\nanover\app\client.py:74, in need_trajectory_joined.<locals>.wrapper(self, *args, **kwargs)
     69 if not self._are_framed_subscribed:
     70     raise RuntimeError(
     71         "You need to first subscribe to the frame using the "
     72         "subscribe_to_frames or the subscribe_to_all_frames methods."
     73     )
---> 74 return func(self, *args, **kwargs)

File ~\anaconda3\envs\subtle-game\lib\site-packages\nanover\app\client.py:411, in NanoverImdClient.current_frame(self)
    402 @property  # type: ignore
    403 @need_frames
[...]

PR #71, opened 2024-04-08
fix copying of frames that contain empty arrays

Ragzouken:
fixes https://github.com/IRL2/nanover-protocol/issues/70

### DIFF
--- a/python-libraries/nanover-core/tests/trajectory/test_frame_data.py
+++ b/python-libraries/nanover-core/tests/trajectory/test_frame_data.py
@@ -737,13 +737,13 @@ def test_copy_doesnt_mutate_original_array(value):
     assert frame.arrays["sample.new"][0] != "baby yoda"
 
 
-@given(ARRAYS_STRATEGIES["string_values"])
-def test_copy_empty_array(value):
+@pytest.mark.parametrize("typed_array", [[0], [0.0], ["string"]])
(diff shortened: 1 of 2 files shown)


python-libraries/nanover-core/tests/trajectory/test_frame_data.py
```diff
@@ -737,6 +737,17 @@ def test_copy_doesnt_mutate_original_array(value):
     assert frame.arrays["sample.new"][0] != "baby yoda"
 
 
+@pytest.mark.parametrize("typed_array", [[0], [0.0], ["string"]])
```
jbarnoud, line 740: I guess this is to make sure the array is created with a type? If so it would be good to have a comment about it.

jbarnoud, line 740: Could `[]` be a valid value to test here? Or will protobuf not let you do that?

Ragzouken: That's a `ValueError` from out `FrameData` class because empty array can't determine type. I've raised this issue for possibilities of supporting it in the future: https://github.com/IRL2/nanover-protocol/issues/72

Ragzouken: Hopefully my added comments clear this up?

PR #74, opened 2024-04-09
Example code / utility for creating a visualisation selection that hides the solvent

Ragzouken:
This is something users actually want to do, so I think it's worth including some simple working example of how to do it. This code works, but I haven't put thought into where it should be or what form it should take (e.g file vs notebook)

### DIFF
--- a/examples/hide_solvent.py
+++ b/examples/hide_solvent.py
@@ -1,6 +1,20 @@
+"""
+Example of using selections to hide solvent by residue name.
+"""
+
 from nanover.app import NanoverImdClient
+from nanover.mdanalysis import frame_data_to_mdanalysis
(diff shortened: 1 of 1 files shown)


jbarnoud:
The example should also show how to deactivate interactions.

jbarnoud (approved):
LGTM

@rhoslynroebuck @hjstroud Does this make sense to you?

hjstroud:
I've just had a quick glance over and I think this looks sensible! I think having it as an example notebook is a good idea too, it's nice to see and actual example of how to implement this sort of feature and I think it is one that will be very useful for a lot of people who would be interested in using VR in their workflows. I'll have a more detailed look at this after the dev meeting!

PR #76, opened 2024-04-09
Fix ESSD C# warnings

Ragzouken:


jbarnoud (approved):
LGTM

jbarnoud:
This should probably make its way to the plugin.

Issue #64, opened 2024-04-04
Remove deprecated `OpenMMIMDRunner`


jbarnoud:
The name `OpenMMIMDRunner` is deprecated in favour of `ASEOpenMMRunner`. This deprecation has had a warning for a very long time by now, we can remove the deprecated symbol.

```
python-libraries/nanover-ase/tests/openmm_ase/test_ase_runner.py: 24 warnings
  /home/jon/narupa/narupa-protocol-gh/python-libraries/nanover-ase/src/nanover/ase/openmm/runner.py:392: DeprecationWarning: The name "OpenMMIMDRunner" is deprecated and will be removed in a later version. Use "ASEOpenMMRunner" instead.
    warnings.warn(
```

PR #77, opened 2024-04-09
remove alias for ASEOpenMMRunner

Ragzouken:
closes https://github.com/IRL2/nanover-protocol/issues/64

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Parallel tests
         run: python -m pytest --cov nanover python-libraries -n auto -m 'not serial'
       - name: Serial tests
-        run: python -m pytest --cov nanover python-libraries -n auto -m 'serial'
+        run: python -m pytest --cov nanover python-libraries -n 1 -m 'serial'
   mypy:
     name: Type analysis for python
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


Ragzouken:
Not sure about the failing tests, they look like port conflicts but it only seems to be affecting this branch atm

python-libraries/nanover-ase/tests/openmm_ase/test_ase_runner.py
```diff
@@ -61,7 +61,7 @@ def logging_params(tmp_path):
     return params
 
 
-@pytest.fixture(params=(ASEOpenMMRunner, OpenMMIMDRunner))
+@pytest.fixture(params=(ASEOpenMMRunner,))
```
jbarnoud, line 64: We do not need the parametrized fixture anymore. We can get rid of `runner_class` and change `runner` to directly call `ASEOpenMMRunner`.

jbarnoud:
Regarding the test, it looks like there is one test that messes up the others; every time we change the tests, we take the risk of changing the failing test. There may be a way to figure out what test does not clean up after itself... Maybe pytest can run a function after tear down to check if the port is in use?

.github/workflows/tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Parallel tests
         run: python -m pytest --cov nanover python-libraries -n auto -m 'not serial'
       - name: Serial tests
-        run: python -m pytest --cov nanover python-libraries -n auto -m 'serial'
+        run: python -m pytest --cov nanover python-libraries -n 1 -m 'serial'
```
Ragzouken, line 29: This is probably our issue: serial tests need to be run in a single process

jbarnoud: Damn copy-paste! It is odd it did not blow up earlier!

jbarnoud:
One last issue I see is that the openmm_graphene notebook shows errors. They are not due to the change from the PR, but this notebook contains many widgets that only are relevant in a live notebook. We should clear the output before merging.

Issue #65, opened 2024-04-04
invalid value encountered in divide


jbarnoud:
Some tests emit the warning `RuntimeWarning: invalid value encountered in divide`. While it is likely to be fine, we should make sure it really is and maybe silence the warning if it is indeed.

```
python-libraries/nanover-core/tests/imd/test_imd_force.py:487
  /home/jon/narupa/narupa-protocol-gh/python-libraries/nanover-core/tests/imd/test_imd_force.py:487: RuntimeWarning: invalid value encountered in divide
    np.subtract(interaction, position)

python-libraries/nanover-core/tests/imd/test_imd_force.py::test_constant_force[position6-interaction6-1-expected_force6]
  /home/jon/narupa/narupa-protocol-gh/python-libraries/nanover-core/src/nanover/imd/imd_force.py:275: RuntimeWarning: invalid value encountered in divide
    force = distance_vector / np.linalg.norm(distance_vector)
```

Ragzouken:
This is the case where the interaction and atom are at the same point. All the other forces drop to zero here, but since this is the constant force I'm not sure what's best. Could make exact zero distance a special case where force drops to zero (is this discontinuity a problem?), or an arbitrary vector is chosen (this is probably what happens in practice, oscillating around the interaction point). Could introduce a drop-off for small distances--or some continuous function to that effect. Not sure what best matches the intent.

Ragzouken:
From meeting: let's go ahead and make it 0 force in this case

PR #78, opened 2024-04-09
Apply 0 force when constant force interaction and particle overlap

Ragzouken:
fixes https://github.com/IRL2/nanover-protocol/issues/65


jbarnoud:
Youhou! The warnings are gone!

PR #80, opened 2024-04-11
Clean up strange autoconnect test

Ragzouken:
May not be the direct cause of any failure, but this test is weird: there's a duplicate fixture, duplicate words in the docstring, and it starts two servers. Some kind of weird merge failure?

### DIFF
--- a/python-libraries/nanover-core/tests/test_autoconnect.py
+++ b/python-libraries/nanover-core/tests/test_autoconnect.py
@@ -106,7 +106,8 @@ def test_autoconnect_separate_servers(broadcastable_servers):
     Tests that an iMD application running on multiple separate servers on multiple ports is discoverable
     and that the client connects to it in the expected way.
     """
-    # Use unique non-default port for discovery
+    # Use unique non-default port for discovery. This avoids interference
+    # with other tests and other servers on the network.
     DISCOVERY_PORT = BROADCAST_PORT + 2
     frame_server, imd_server, multiplayer_server = broadcastable_servers
 
--- a/python-libraries/nanover-core/tests/test_autoconnect.py
+++ b/python-libraries/nanover-core/tests/test_autoconnect.py
@@ -8,6 +8,7 @@ from nanover.app import NanoverImdApplication, NanoverImdClient
 from nanover.app.app_server import MULTIPLAYER_SERVICE_NAME, DEFAULT_NANOVER_PORT
 from nanover.core import NanoverServer
 from nanover.essd import DiscoveryServer, ServiceHub
+from nanover.essd.server import BROADCAST_PORT
 from nanover.essd.utils import get_broadcastable_ip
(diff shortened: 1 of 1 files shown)


python-libraries/nanover-core/tests/test_autoconnect.py
```diff
@@ -46,34 +46,23 @@ def discoverable_imd_server():
         yield app_server
 
 
-@pytest.fixture
-def discoverable_imd_server():
-    """
-    Returns a discoverable iMD server discoverable on the free port.
-    """
-    DISCOVERY_PORT = 39421
```
jbarnoud, line 54: It is curious that this is not the default ESSD port. Maybe we did that to avoid interference with other servers on the network?

Ragzouken: I was also unsure about this, but it does looks like ESSD has no behavior for picking any available port (which makes sense, it's supposed to run on a default port so you can find it), so it makes sense to pick specific non-overlapping ports for testing them in parallel. Maybe there's a more explicit way to pick sequential port numbers in a fixture?

jbarnoud: Sequential ports would be ideal, but the non-default port is likely enough. (Would be worth documenting the reasoning, though)

Ragzouken: We have to use explicit non-default ports at least, because broadcast doesn't have the means to find an unused one. Are these changes sufficent?

python-libraries/nanover-core/tests/test_autoconnect.py
```diff
@@ -115,7 +106,8 @@ def test_autoconnect_separate_servers(broadcastable_servers):
     Tests that an iMD application running on multiple separate servers on multiple ports is discoverable
     and that the client connects to it in the expected way.
     """
-    DISCOVERY_PORT = 39423
+    # Use unique non-default port for discovery
```
jbarnoud, line 109:
```suggestion
    # Use unique non-default port for discovery. This avoids interference
    # with other tests and other servers on the network.
```

python-libraries/nanover-core/tests/test_autoconnect.py
```diff
@@ -36,22 +37,10 @@ def broadcastable_servers():
...
-        yield app_server
-
-
-@pytest.fixture
-def discoverable_imd_server():
-    """
-    Returns a discoverable iMD server discoverable on the free port.
-    """
-    DISCOVERY_PORT = 39421
+    # Use unique non-default port for discovery
```
jbarnoud, line 42:
```suggestion
    # Use unique non-default port for discovery. This avoids interference
    # with other tests and other servers on the network.
```

jbarnoud (approved):

PR #87, opened 2024-04-25
Update the README

### LINKED ISSUE #85: Specify the required version of dotnet in the README
On Windows, it seems that dotnet can find the version of the runtime and SDK it needs. On Linux and MacOS, the old version we require does not install automatically. We need to write down in the documentation what version to install.

### LINKED ISSUE #86: Document the `--no-dotnet` and `--no-python` options of the compile script
In the README, we should tell about the `--no-dotnet` option of the compile script. Most users will not need to compile the C# code, and it is difficult to compile anyway because of how old of a dotnet version we need (see #85).

We should also document the `--no-python` option.

jbarnoud:
* document the --no-dotnet option for linux and macos (fixes #86)
* document the required version of dotnet (fixes #85)
* stop mentionning the omnia conda repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,13 +97,17 @@ Developers will want the manual install described below.
 ### Setup nanover-protocol for developers on Mac and Linux
 
 * Install Anaconda
-* Install dotnet
 * Clone the nanover-protocol repository
 * In a terminal, in the repository root:
     * Create a conda environment (here we call the environment "nanover-dev"): `conda create -n nanover-dev "python>3.11"`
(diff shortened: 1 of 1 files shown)


jbarnoud:
@hjstroud could you have a look? Does it make sense to you?

hjstroud:
> @hjstroud could you have a look? Does it make sense to you?

I'll take a look now!

hjstroud (approved):
All good, go for it!

Issue #46, opened 2024-02-27
intermittent failure on `test_imd_client.py::test_interactions_property`


Ragzouken:
This test failed on an unrelated change and worked on rerun. Guessing it's a timing issue or just a freak accident--logging it here in case it happens again.

```
=========================== short test summary info ============================
FAILED python-libraries/nanover-core/tests/imd/test_imd_client.py::test_interactions_property - AssertionError: assert dict_keys([]) == dict_keys(['i.....second_id'])
  
  Full diff:
  - dict_keys(['interaction..first_id', 'interaction..second_id'])
  + dict_keys([])
============ 1 failed, 796 passed, 60 warnings in 112.52s (0:01:52) ============
Error: Process completed with exit code 1.
```

Ragzouken:
Looks like it didn't wait long enough for the server to respond. It'd be convenient if we could come up with a clean and easy utility for doing something of the form `assert_x_within_timeout(timeout, lambda: x)`

PR #112, opened 2024-05-08
pass-early delayed assert

Ragzouken:
We have a lot of tests that depend on waiting for threaded/network activity to happen in a timely fashion. The longer we wait before asserting, the more reliable the test, but the longer the entire suite will take. This is a possible pattern we could use to only wait as long as necessary.

It would require corresponding functions for not equals, in, not in, etc. Not sure if it could be made any cleaner. Also not sure how to make a module available to the tests without just putting it in `nanover.core`.

This might just be a bad approach, but I present it here as an option to discuss.

Closes https://github.com/IRL2/nanover-protocol/issues/46

### DIFF
--- a/python-libraries/nanover-core/tests/imd/test_imd_client.py
+++ b/python-libraries/nanover-core/tests/imd/test_imd_client.py
@@ -1,5 +1,7 @@
 import itertools
 import time
+from typing import Callable
+
 import pytest
 from nanover.imd.particle_interaction import ParticleInteraction
 from nanover.imd.imd_state import (
@@ -12,6 +14,14 @@ from nanover.utilities.change_buffers import DictionaryChange
 from .test_imd_server import imd_server_client, imd_server, interaction
 
 IMMEDIATE_REPLY_WAIT_TIME = 0.01
+
+
+def assert_equal_soon(a: Callable, b: Callable, interval=0.1, timeout=1.0):
+    __tracebackhide__ = True
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline and a() != b():
+        time.sleep(interval)
+    assert a() == b()
 
 
 def test_start_interaction(imd_server_client):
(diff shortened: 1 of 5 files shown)


python-libraries/nanover-core/tests/imd/test_imd_client.py
```diff
@@ -182,6 +192,6 @@ def test_interactions_property(imd_server_client):
     imd_server.update_state(None, interaction_updates)
 
     imd_client.subscribe_all_state_updates(interval=0)
-    time.sleep(IMMEDIATE_REPLY_WAIT_TIME)
-    print(imd_client.interactions)
-    assert imd_client.interactions.keys() == real_interactions.keys()
+    assert_equal_soon(
+        lambda: imd_client.interactions.keys(), lambda: real_interactions.keys()
```
jbarnoud, line 196:
```suggestion
        imd_client.interactions.keys, real_interactions.keys
```

jbarnoud (commented):
Good idea.

jbarnoud:
> Not sure how to go about making this available to all the testing files via import, is there an easy way?

You could create `nanover.core.testing` and import it wherever you need it.

python-libraries/nanover-core/tests/imd/test_imd_client.py
```diff
@@ -14,6 +16,30 @@
 IMMEDIATE_REPLY_WAIT_TIME = 0.01
 
 
+def assert_equal_soon(a: Callable, b: Callable, interval=0.1, timeout=1.0):
+    __tracebackhide__ = True
```
jbarnoud, line 20: This is some pytest magic I did not know about. It may be good to add a comment explaining that it makes pytest hide the function in the traceback, maybe link to the pytest doc, and explaining the rationale of the hiding here.

Does the function appearing in the traceback really make a difference, though?

jbarnoud:
I really like the name of the assertion functions!

python-libraries/nanover-core/tests/imd/test_imd_client.py
Ragzouken: this is the difference:
image
image

Issue #120, opened 2024-05-13
Throttling tests failing in python-libraries/nanover-openmm/tests_nanover_openmm/test_runner.py


hjstroud:
There is a test (/are multiple tests?) in python-libraries/nanover-openmm/tests_nanover_openmm/test_runner.py that seems to fail consistently when running all of the tests, but doesn't fail when running that test file in isolation (at least this appears to be the behaviour on Mac). When it fails, it does so with messages like

FAILED python-libraries/nanover-openmm/tests_nanover_openmm/test_runner.py::TestRunner::test_throttling[1-5] - assert 0.20507699400020232 == 0.2 ± 5.0e-03

or

FAILED python-libraries/nanover-openmm/tests_nanover_openmm/test_runner.py::TestRunner::test_throttling[5-30] - assert 0.03849816671364741 == 0.03333333333333333 ± 5.0e-03

hjstroud:
UPDATE: This does also seem to occur when running the tests in this file in isolation

PR #121, opened 2024-05-13
More fixes for throttling tests + isolate timing sensitive tests

Ragzouken:
Fixes https://github.com/IRL2/nanover-protocol/issues/120

We can't really guarantee precise timing if the system is under too much strain, so the tests can reflect that by running the tests in isolation from other potentially heavy tests running in parallel. 

There was also a bug that repeatedly did `_on_first_frame` which lead to repeated frame sends for clients with no subscribe interval.

I've also relaxed the timing tolerance to 7ms for throttling since it seems to fail easily for 5ms even after these changes.

### DIFF
--- a/python-libraries/nanover-ase/tests/openmm_ase/test_ase_runner.py
+++ b/python-libraries/nanover-ase/tests/openmm_ase/test_ase_runner.py
@@ -22,7 +22,7 @@ from nanover.ase.wall_constraint import VelocityWallConstraint
 from .simulation_utils import basic_simulation, serialized_simulation_path
 
 TRAJECTORY_OUTPUT_FILENAME = "test.xyz"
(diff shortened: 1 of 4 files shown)


hjstroud:
I think this all looks sensible to me! My only comment is about the number of frames used for the test: is 6 frames enough? Other than this everything looks nice and clear on the face of it!

Ragzouken:
> I think this all looks sensible to me! My only comment is about the number of frames used for the test: is 6 frames enough? Other than this everything looks nice and clear on the face of it!

6 is probably not enough, but I was keeping it low so the 1fps test didn't take any longer than 6s to run. But it's probably more sensible to just drop the 1fps case and up the test frame count, so I've done that now.

Issue #33, opened 2023-10-19
Windows LAMMPS tests fail with `Windows fatal exception: access violation`


jbarnoud:

Ragzouken:
Can we get more info in here on how to reproduce this? Looks like something to do with a lammps test?

jbarnoud:
This has not occurred in a while and I do not remember what the issue was exactly. Closing for now.

jbarnoud:
Nevermind, it did not occur anymore because I skipped the tests... We should remove the skip to see if the issue is still there.

Ragzouken:
It finally returned:

```
Run python -m pytest --cov nanover python-libraries -n auto -m 'not serial'
============================= test session starts =============================
platform win32 -- Python 3.12.6, pytest-[8](https://github.com/IRL2/nanover-protocol/actions/runs/11052367776/job/30704412671?pr=244#step:7:9).3.3, pluggy-1.5.0
rootdir: D:\a\nanover-protocol\nanover-protocol
configfile: pytest.ini
plugins: anyio-4.6.0, hypothesis-6.112.1, cov-5.0.0, timeout-2.3.1, xdist-3.6.1
created: 2/2 workers
2 workers [818 items]

........................................................................ [  8%]
........................................................................ [ 17%]
........................................................................ [ 26%]
........................................................................ [ 35%]
........................................................................ [ 44%]
........................................................................ [ 52%]
........................................................................ [ 61%]
........................................................................ [ 70%]
Windows fatal exception: access violation

Thread 0x000011d4 (most recent call first):
  File "C:\Miniconda3\envs\test\Lib\site-packages\execnet\gateway_base.py", line 534 in read
  File "C:\Miniconda3\envs\test\Lib\site-packages\execnet\gateway_base.py", line 567 in from_io
  File "C:\Miniconda3\envs\test\Lib\site-packages\execnet\gateway_base.py", line 1160 in _thread_receiver
  File "C:\Miniconda3\envs\test\Lib\site-packages\execnet\gateway_base.py", line 341 in run
  File "C:\Miniconda3\envs\test\Lib\site-packages\execnet\gateway_base.py", line 411 in _perform_spawn

Current thread 0x00001[9](https://github.com/IRL2/nanover-protocol/actions/runs/11052367776/job/30704412671?pr=244#step:7:10)28 (most recent call first):
  File "<frozen importlib._bootstrap>", line 488 in _call_with_frames_removed
[...]

PR #122, opened 2024-05-13
Restore lammps tests

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/33 or at least attempts to make it a problem again.


jbarnoud:
Let's hope this does the trick. Otherwise, the issue can always be reopened.

Issue #31, opened 2023-10-19
CI fails systematically


jbarnoud:
All jobs in CI fail when executing `conda install`. Is this a transient error due to an external factor or is this an issue with our CI?

In practice, `conda install` seems to succeed, but ends with an exit code of 1 indicating failure.

jbarnoud:
The issue is hidden behind a workaround. We would need to remove the workaround to see if it is fixed.

Ragzouken:
maybe it's this:
![image](https://github.com/IRL2/nanover-protocol/assets/527158/05df47ea-988d-4e49-a536-33d8ac525873)

PR #126, opened 2024-05-15
Don't suppress CI failure for conda install

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/31 until it happens again


jbarnoud (approved):
We'll see. Hopefully it was a transcient problem and we will never have to think about it again.

Issue #117, opened 2024-05-10
Run CI for mac


Ragzouken:
We ran into a platform specific issue already and Mac use is not uncommon among our intended users, I think.

PR #127, opened 2024-05-15
Run CI on Mac

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/117


Ragzouken:
The available CI Mac seems to run slower, so either we extend all the sleeps in the code or replace them with something like this: https://github.com/IRL2/nanover-protocol/pull/112

Ragzouken:
Although even the interval stuff seems to be way off so maybe something is really wrong..

PR #134, opened 2024-05-20
Implement simulation switching for omm-ase

Ragzouken:


Ragzouken:
I'll repeat this for the bare OMM runner in another request.

python-libraries/nanover-ase/src/nanover/ase/openmm/runner.py
```diff
@@ -188,6 +206,26 @@ def reset(self):
     def step(self):
         self.imd.step()
 
+    def load(self, index: int):
+        self._sim_index = int(index % len(self.simulation_list))
```
jbarnoud, line 210: This needs to be documented: indices cycle when using both `load` and `next`.

python-libraries/nanover-ase/src/nanover/ase/openmm/runner.py
```diff
@@ -188,6 +206,26 @@ def reset(self):
...
+        self.simulation = self.simulation_list[self._sim_index]
+        self._initialise_calculator(self.simulation, walls=self._imd_params.walls)
+        self._initialise_dynamics()
+        self.imd.replace_dynamics(self.dynamics)
+        self.imd.play()
+
+    def next(self):
+        self.load(self._sim_index + 1)
+
+    def list(self):
```
jbarnoud, line 220: In the rust server I *try* to limit how much I reveal about the server's file system. I manage poorly but telling too much about the structure of the file system may be an issue in a cloud context.

https://github.com/IRL2/nanover-rs/blob/cf86bf1da2ec037dc73b7ec96d169c60ec839ca0/nanover-rs/src/playback.rs#L151

python-libraries/nanover-ase/src/nanover/ase/imd.py
```diff
@@ -311,6 +307,38 @@ def reset(self):
         self.atoms.set_cell(self._initial_box)
         self._call_on_reset()
 
+    def replace_dynamics(self, dynamics: MolecularDynamics, reset=True):
+        if reset:
```
jbarnoud, line 311: Do we document the reset event anywhere? The `on_reset_listener`? It should document that when a new simulation is loaded, the event is called from the old simulation, based on where the reset call is located.

jbarnoud (commented):
This looks sound.

@rhoslynroebuck could you try running the SublteGame from this server? It is not yet suitable for your study but it should be able to run the game.

python-libraries/nanover-ase/src/nanover/ase/openmm/runner.py
jbarnoud: Should be fine indeed. It may not be super informative for some users that may call all their files `simulation.xml` but that's more robust than my attempt.

python-libraries/nanover-ase/src/nanover/ase/imd.py
Ragzouken: Not sure. I don't know what it was intended for, but it seems it happens either when the user chooses to reset or when the simulation explodes and resets itself. The two places it's used are in the CLI to print loudly that a reset happened, and in the trajectory logger where resets start a new recording file.

BTW do you see any reason why there should be a difference between a reset and reloading the current simulation?

python-libraries/nanover-ase/src/nanover/ase/openmm/runner.py
Ragzouken: Do you mean in a comment here, a docstring here, or somewhere else? I could also just change it to do whatever the rust server does?

python-libraries/nanover-ase/src/nanover/ase/imd.py
jbarnoud: If we just loudly yell reset we should be good whatever simulation is current when the event occurs.

There is a small semantic difference between reset and reload. When resetting, you do not expect the topology to change while you expect it when loading a new simulation. Sending a new topology will affect `nanover.mdanalysis` when reading recordings since it stops whenever it receives a new topology.

rhoslynroebuck:
I just tried this with subtle game and I found that I'm not receiving the `system.simulation.counter` key:

```
Traceback (most recent call last):
  File "C:\Users\rhosl\GIT_REPOS\SubtleGame\Client\puppeteering_client.py", line 223, in <module>
    puppeteering_client.run_game()
  File "C:\Users\rhosl\GIT_REPOS\SubtleGame\Client\puppeteering_client.py", line 98, in run_game
    simulation_counter = self.nanover_client.current_frame.values["system.simulation.counter"]
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rhosl\miniforge3\envs\subtle-game\Lib\site-packages\nanover\trajectory\frame_data.py", line 433, in __getitem__
    raise KeyError(f'No {self.singular} with the key "{key}".')
KeyError: 'No value with the key "system.simulation.counter".'
```

Ragzouken:
> I just tried this with subtle game and I found that I'm not receiving the `system.simulation.counter` key:
> 
> ```
> Traceback (most recent call last):
>   File "C:\Users\rhosl\GIT_REPOS\SubtleGame\Client\puppeteering_client.py", line 223, in <module>
>     puppeteering_client.run_game()
>   File "C:\Users\rhosl\GIT_REPOS\SubtleGame\Client\puppeteering_client.py", line 98, in run_game
>     simulation_counter = self.nanover_client.current_frame.values["system.simulation.counter"]
>                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "C:\Users\rhosl\miniforge3\envs\subtle-game\Lib\site-packages\nanover\trajectory\frame_data.py", line 433, in __getitem__
>     raise KeyError(f'No {self.singular} with the key "{key}".')
> KeyError: 'No value with the key "system.simulation.counter".'
> ```

I had the wrong name for this key, the current commit should fix it. Is counter supposed to start at 0 or 1 when the starting simulation is running?

rhoslynroebuck:
I now receive the following error:

```
  File "c:\users\rhosl\git_repos\nanover-protocol\python-libraries\nanover-ase\src\nanover\ase\imd.py", line 332, in replace_dynamics
    self.atoms, self._frame_publisher, self._simulation_counter - 1
                                       ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NanoverASEDynamics' object has no attribute '_simulation_counter'. Did you mean: '_simulation_count'?
```

jbarnoud:
The counter starts at 0.

Ragzouken:
> I now receive the following error:
> 
> ```
>   File "c:\users\rhosl\git_repos\nanover-protocol\python-libraries\nanover-ase\src\nanover\ase\imd.py", line 332, in replace_dynamics
>     self.atoms, self._frame_publisher, self._simulation_counter - 1
>                                        ^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NanoverASEDynamics' object has no attribute '_simulation_counter'. Did you mean: '_simulation_count'?
> ```

Sorry, I messed it up again. Should work now if you could try one more time.

rhoslynroebuck:
This is all good now!

Issue #44, opened 2024-02-26
Read shared state recordings


jbarnoud:
The rust server and nanover-recorder record not only the trajectory but also the shared state. While we can read the trajectory with `nanover-mdanalysis`, we cannot read the shared state recording yet. Here is a quick attempt at the task:

```python
#!/usr/bin/env python

"""
Read the shared state time series as recorded by Narupa.

This file can be used as a python module or as a script.

Example when used as a module:

.. code:: python

    import read_state

    # Print the timestamps and the state updates
    from timestamp, update in iter_state_file('my_file.state'):
        print(timestamp, update)

    # Print the timestamps and the full state
    from timestamp, state in iter_full_states(iter_state_file('my_file.state')):
        print(timestamp, state)

Example when used as a script:

.. code:: bash

    # show the help
    python read_state.py --help
    
    # show the timestamps and state updates one record per line
    python read_state.py my_file.state
    # show the timestamps and the state updates in human-readable form
    python read_state.py --pretty my_file.state
    
    # show the timestamps and full states one record per line
    python read_state.py --full my_file.state
    # show the timestamps and full state in human-readable form
    python read_state.py --full --pretty my_file.state

"""

import argparse
from pprint import pprint
from narupa.mdanalysis import recordings
from narupa.protocol.state import StateUpdate
from narupa.utilities.protobuf_utilities import struct_to_dict

def iter_state_updates(unpacker: recordings.Unpacker) :
    """
    Read a binary stream and yield the state updates and their timestamps.
    """
    supported_format_versions = (2,)
    magic_number = unpacker.unpack_u64()
    if magic_number != recordings.MAGIC_NUMBER:
        raise recordings.InvalidMagicNumber
    format_version = unpacker.unpack_u64()
    if format_version not in supported_format_versions:
        raise recordings.UnsuportedFormatVersion(format_version, supported_format_versions)
    while True:
        try:
            elapsed = unpacker.unpack_u128()
            record_size = unpacker.unpack_u64()
            buffer = unpacker.unpack_bytes(record_size)
        except IndexError:
            break
        state_update = StateUpdate()
        state_update.ParseFromString(buffer)
        yield (elapsed, struct_to_dict(state_update.changed_keys))


def iter_full_states(updates):
    """
    Read a stream of timestamps and state updates and yield the timestamp and the aggregated state.
    """
    aggregate_state = {}
    for timestamp, update in updates:
        aggregate_state.update(update)
[...]

PR #152, opened 2024-06-25
Improve traj/state reading/writing

Ragzouken:
Need some tests ideally..


Ragzouken:
closes https://github.com/IRL2/nanover-protocol/issues/44

Ragzouken:
Found some strange problems reading the example state file.. need to figure it out and put a test in

hjstroud (approved):
All good to me! Went through it online with @Ragzouken and all seems to be in order, nice one!

PR #158, opened 2024-07-03
Fix wrapping of OMM positions

Ragzouken:
Don't accidentally wrap OpenMM positions contrary to what comments say is happening


hjstroud (approved):
Perfect, this should fix the weird periodic behaviour I was observing!

PR #165, opened 2024-07-05
Output sparse arrays of user forces and corresponding indices

hjstroud:
This is a PR to add sparse arrays of user forces and indices to the output of NanoVer simulations. The user forces and indices are labelled the same way in FrameData as they are in the Rust server. This means that they should also be readable by the mdanalysis functions that already exist to parse sparse user forces into full arrays of user forces (python server recording functionality required to test this).

### DIFF
--- a/python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
+++ b/python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
@@ -299,7 +299,7 @@ class FrameData(metaclass=_FrameDataMeta):
         to_python=_n_by_3,
         to_raw=_flatten_array,
     )
-    sparse_user_index: float = _Shortcut(  # type: ignore[assignment]
+    sparse_user_index: List[float] = _Shortcut(  # type: ignore[assignment]
         key=SPARSE_USER_INDEX,
         record_type="arrays",
         field_type="float",
--- a/python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
+++ b/python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
@@ -50,6 +50,7 @@ _Shortcut = namedtuple(
 
 Array2Dfloat = Union[List[List[float]], npt.NDArray[Union[np.float32, np.float64]]]
 Array2Dint = Union[List[List[int]], npt.NDArray[Union[np.int_]]]
+Array1Dint = Union[List[int], npt.NDArray[np.int_]]
 
 
 class MissingDataError(KeyError):
@@ -299,12 +300,12 @@ class FrameData(metaclass=_FrameDataMeta):
         to_python=_n_by_3,
         to_raw=_flatten_array,
     )
-    sparse_user_index: List[float] = _Shortcut(  # type: ignore[assignment]
+    sparse_user_index: Array1Dint = _Shortcut(  # type: ignore[assignment]
         key=USER_FORCES_INDEX,
         record_type="arrays",
-        field_type="float",
-        to_python=_n_by_3,
-        to_raw=_flatten_array,
+        field_type="index",
+        to_python=_as_is,
+        to_raw=_as_is,
     )
 
     box_vectors: List[List[float]] = _Shortcut(  # type: ignore[assignment]
--- a/python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
+++ b/python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
@@ -38,8 +38,8 @@ POTENTIAL_ENERGY = "energy.potential"
 TOTAL_ENERGY = "energy.total"
 USER_ENERGY = "energy.user.total"
 
-SPARSE_USER_FORCES = "forces.user.sparse"
-SPARSE_USER_INDEX = "forces.user.index"
+USER_FORCES_SPARSE = "forces.user.sparse"
+USER_FORCES_INDEX = "forces.user.index"
 
 SERVER_TIMESTAMP = "server.timestamp"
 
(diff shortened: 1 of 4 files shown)


python-libraries/nanover-openmm/src/nanover/openmm/imd.py
```diff
@@ -114,6 +114,7 @@ def __init__(
         self._previous_force_index: Set[int] = set()
         self._frame_index = 0
         self._total_user_energy = 0.0
+        self._user_forces = None
```
Ragzouken, line 117: You'll need something like this to satisfy the type checker:
```suggestion
        self._user_forces: Optional[npt.NDArray] = None
```

python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
```diff
@@ -289,6 +292,20 @@ class FrameData(metaclass=_FrameDataMeta):
...
         to_raw=_as_is,
     )
+    sparse_user_forces: Array2Dfloat = _Shortcut(  # type: ignore[assignment]
+        key=SPARSE_USER_FORCES,
+        record_type="arrays",
+        field_type="float",
+        to_python=_n_by_3,
+        to_raw=_flatten_array,
+    )
+    sparse_user_index: float = _Shortcut(  # type: ignore[assignment]
```
Ragzouken, line 302: This is an array of float values, not a single float:
```suggestion
    sparse_user_index: List[float] = _Shortcut(  # type: ignore[assignment]
```

Ragzouken:
Seem to have a repeatable test failure: `pytest python-libraries/nanover-openmm/tests_nanover_openmm/test_imd.py::TestNanoverImdReporter::test_interactions_interval`

Ragzouken:
We'll want to add a test at least to verify that given some input forces that the right frame contains the expected output

python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
```diff
@@ -38,6 +38,9 @@
 TOTAL_ENERGY = "energy.total"
 USER_ENERGY = "energy.user.total"
 
+SPARSE_USER_FORCES = "forces.user.sparse"
+SPARSE_USER_INDEX = "forces.user.index"
```
Ragzouken, line 42: let's match the names of fields more closely as:
`USER_FORCES_SPARSE`
`USER_FORCES_INDEX`

python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
hjstroud: Good spot!

python-libraries/nanover-openmm/src/nanover/openmm/imd.py
hjstroud: Another good spot!

python-libraries/nanover-openmm/src/nanover/openmm/imd.py
```diff
@@ -171,6 +172,9 @@ def report(self, simulation: app.Simulation, state: mm.State) -> None:
             else:
                 frame_data.particle_positions = positions
                 frame_data.user_energy = self._total_user_energy
+                sparse_indices, sparse_forces = get_sparse_forces(self._user_forces)
```
Ragzouken, line 175: to satisfy the type checker you need to check that `self._user_forces` is not `None` before trying to use it, but I suggest if it's valid to instead not allow `self._user_forces` to be `None` and initialise it as an empty array to begin with

python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
hjstroud: All sorted

python-libraries/nanover-openmm/src/nanover/openmm/imd.py
```diff
@@ -357,3 +362,13 @@ def get_imd_forces_from_system(system: app.Simulation) -> List[mm.CustomExternal
         and force.getEnergyFunction() == IMD_FORCE_EXPRESSION
         and force.getNumParticles() == system_num_particles
     ]
+
+
+def get_sparse_forces(forces: npt.NDArray) -> Tuple[npt.NDArray, npt.NDArray]:
```
Ragzouken, line 367: Could do with a docstring here to be explicit about the meaning of sparse forces i.e that it's a pair of lists of forces and the corresponding particle index (I think?)

python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
```diff
@@ -289,6 +293,20 @@ class FrameData(metaclass=_FrameDataMeta):
...
+        to_python=_n_by_3,
+        to_raw=_flatten_array,
+    )
+    sparse_user_index: Array1Dint = _Shortcut(  # type: ignore[assignment]
+        key=USER_FORCES_INDEX,
+        record_type="arrays",
+        field_type="index",
+        to_python=_as_is,
+        to_raw=_as_is,
+    )
```
Ragzouken, line 309: Should rename these to match the renamed constants i.e `user_forces_sparse` and `user_forces_index`

python-libraries/nanover-openmm/tests_nanover_openmm/test_imd.py
```diff
@@ -185,6 +185,24 @@ def test_report_send_first_frame(self, app_simulation_and_reporter):
...
+        request_id = app.frame_publisher._get_new_request_id()
+        frame_queues = app.frame_publisher.frame_queues
+        with frame_queues.one_queue(request_id, Queue) as publisher_queue:
+            simulation.step(10)
+            frames = list(publisher_queue.queue)
+
+        # frame 0: topology special case
+        # frame 1: no previous frame so no forces?
+        # frame 2: forces from frame 1?
+        frame = FrameData(frames[2].frame)
```
Ragzouken, line 202: I put this here because I'm not entirely sure how it's supposed to work and we should be clear on which frames are expected to contain user forces and whether they appear before or after the force actually takes affect

hjstroud: Fantastic, this looks like a sensible check to me.

python-libraries/nanover-openmm/src/nanover/openmm/imd.py
hjstroud: I have since added a docstring to this function, let me know if it needs further editing!

python-libraries/nanover-core/src/nanover/trajectory/frame_data.py
hjstroud: These have now been renamed as per this comment

Ragzouken (approved):
Looks good to me. Though a test for specific values for the reported force wouldn't hurt, if it's viable.

PR #191, opened 2024-07-24
Replace old CLIs with notice to use `nanover-omni` instead

Ragzouken:
I think the docs already no longer mention these two CLIs, and I think that everyone is already either using the runners via notebooks or using the rust server instead


Mohamed-Dhouioui (approved):
If no one else needs this, I think what you did is reasonable.

hjstroud (approved):
This looks sensible to me. As long as no-one else is using it I'm happy with this!

Issue #196, opened 2024-07-24
OpenMM Omni runner won't run without explicitly adding IMD force in example `OpenMM runner` Jupyter notebook


hjstroud:
We tested this notebook by commenting out the `add_imd_force_to_system` line and the OpenMM runner wouldn't work, throwing the error 

`[nanover-protocol/python-libraries/nanover-omni/src/nanover/omni/openmm.py:25](http://localhost:8888/lab/workspaces/~/IRL/nanover-protocol/python-libraries/nanover-omni/src/nanover/omni/openmm.py#line=24), in OpenMMSimulation.from_simulation(cls, simulation, name)
     23 potential_imd_forces = get_imd_forces_from_system(sim.simulation.system)
     24 if not potential_imd_forces:
---> 25     raise ValueError(
     26         "The simulation must include an appropriate force for imd."
     27     )
     28 if len(potential_imd_forces) > 1:
     29     warnings.warn(
     30         f"More than one force could be used as imd force "
     31         f"({len(potential_imd_forces)}); taking the last one."
     32     )

ValueError: The simulation must include an appropriate force for imd.`

We then tried running the same thing but exchanging the OpenMM runner for the Omni runner, and ran into the same error.

However, when we ran the same simulation file (without defining a imd force) using the command line interface Omni runner, it worked and the simulation could be interacted with. @Ragzouken what's going on, is this a bug or a misunderstanding on our part?

Ragzouken:
The reason behind all this is that adding forces after the system is already set up won't actually get properly accounted for--so it's necessary to add them as the simulation is constructed. Any time you load via the dedicated XML methods (including the command line) this is possible because the runner is in charge of constructing the simulation. If instead you construct the simulation manually, then it needs to be added manually at that stage because by the time you give that simulation to the runner then it's already too late.

I did look into it yesterday and find that there might be a way to update the forces after the fact. Here's a merge request: https://github.com/IRL2/nanover-protocol/pull/199 if you could test if the interactions really work, then we could switch to that and save a lot of trouble and never have to add force manually.

Ragzouken:
As of most recent merge, you shouldn't need to add the force manually for omni's `OpenMMSimulation`

PR #199, opened 2024-07-24
add imd force automatically in omni's omm 

Ragzouken:
May fix https://github.com/IRL2/nanover-protocol/issues/196


Ragzouken:
should probably figure out an automated interaction test to go along with this if it really does work

PR #201, opened 2024-07-29
Optionally add particle velocities and forces to FrameData

### LINKED ISSUE #37: Optionally include forces and velocities to the FrameData
The rust server can *optionally* include the forces and velocities for each particle in the FrameData under the `particle.forces` and the `particle.velocities` keys, respectively. This allows client to access this data, but mostly it allows to record them for later analysis. This should be included in the runners.

* [ ] OpenMM runner
* [ ] ASE runner
* [ ] LAMMPS runner

hjstroud:
This PR aims to fix #37 , by giving the user the option to include the velocities and (total) forces acting on the particles when using OpenMM. This fix should enable the user to optionally include the forces, velocities or both quantities in the FrameData delivered to the client.


python-libraries/nanover-openmm/src/nanover/openmm/cli.py
```diff
@@ -63,6 +63,20 @@ def handle_user_arguments() -> argparse.Namespace:
...
+        default=False,
+        help="Optionally include the particle velocities in the frame data.",
+    )
+    parser.add_argument(
+        "-k",
+        "--include-forces",
+        type=bool,
+        default=False,
+        help="Optionally include the particle forces in the frame data.",
+    )
```
Ragzouken, line 79: These should be `action="store_true"` with no `type=bool` -- otherwise it expects a value to be provided

python-libraries/nanover-openmm/src/nanover/openmm/cli.py
```diff
@@ -63,6 +63,20 @@ def handle_user_arguments() -> argparse.Namespace:
         metavar="STEPS",
         help="Update the interactions every STEPS dynamics steps.",
     )
+    parser.add_argument(
+        "-q",
+        "--include-velocities",
+        type=bool,
```
hjstroud, line 69: Replace `type=bool` with
```suggestion
        action="store-true",
```

python-libraries/nanover-openmm/src/nanover/openmm/cli.py
```diff
@@ -63,6 +63,20 @@ def handle_user_arguments() -> argparse.Namespace:
...
+        default=False,
+        help="Optionally include the particle velocities in the frame data.",
+    )
+    parser.add_argument(
+        "-k",
+        "--include-forces",
+        type=bool,
```
hjstroud, line 76: Replace `type=bool` with
```suggestion
        action="store_true",
```

python-libraries/nanover-openmm/src/nanover/openmm/cli.py
hjstroud: Added this change, thank you for spotting this!

Ragzouken (approved):
Looks good!

Ragzouken (approved):
looks good!

PR #205, opened 2024-08-05
Define selections per simulation on OmniRunner

Ragzouken:
image


examples/omni_multiple.ipynb
```diff
@@ -0,0 +1,253 @@
...
+    "    omni_runner.close()\n",
+    "except NameError:\n",
+    "    pass\n",
+    "omni_runner = OmniRunner.with_basic_server(port=0)\n",
+    "\n",
+    "omni_runner.add_simulation(omni_nanotube)\n",
+    "omni_runner.set_simulation_selections(\n",
+    "    omni_nanotube, \n",
+    "    nanotube_selection, \n",
+    "    make_solvent_hiding_selection(omni_nanotube)\n",
```
hjstroud, line 165: The nanotube system has no solvent, so this throws an error for me when I hit `.next()`

examples/omni_multiple.ipynb
```diff
@@ -0,0 +1,253 @@
...
+    ")\n",
+    "\n",
+    "omni_runner.add_simulation(omni_alanine)\n",
+    "omni_runner.set_simulation_selections(omni_alanine, alanine_selection)\n",
+    "\n",
+    "omni_runner.next()\n",
```
hjstroud, line 171: `.next()` seems to immediately load the second simulation, rather than loading the first one.

Is it also worth adding a cell with `omni_runner.next()` to allow the client to run the cell multiple times to see the simulations switching between the two loaded simulations with the pre-defined selections?

hjstroud (changes requested):
This mostly looks good to me, and with a couple of minor changes the notebook ran fine and produced the expected results. As far as changes go we've chatted about this in person, but for reference I would request the following:

- Removing the "hide solvent" selection from the nanotube selections, because the nanotube file doesn't contain water (there is no solvent)
- Changing the loading of the first simulation to use `.load(0)` (with a comment?) rather than `.next()`, which moves straight to the next simulation
- Adding a cell that has `omni_runner.next()` in, with an accompanying markdown cell to encourage the user to rerun this cell to switch between simulations

Other than these minor things, it all looks good, nice work!

Ragzouken:
> One more thing! The current location of the notebook means that the file paths are invalid: is this example notebook supposed to be in the OpenMM directory? If not, the file path needs updating. I'm happy for this notebook to not yet have an assigned directory (like the hide solvent script), and to just live in the base examples directory for now.

Ah yes, I have corrected the paths assuming that it will stay where it is for now.

hjstroud (approved):
Changes look good to me. I have added a cell with the `.next()` function and a comment for the users. As long as you're happy with this, I approve this PR. Nice work!

ludovicaisa:
PR approved

PR #207, opened 2024-08-07
Exclude interaction energy from potential energy

Ragzouken:


hjstroud (approved):
This is a great way to go about fixing this issue. We have chatted about it fairly extensively, and I'm happy for this to go in, as it appears to fix the issue of incorrect potential energies (at least as far as I can see)

PR #216, opened 2024-08-12
Add OpenMM tutorial notebooks

hjstroud:
This PR fleshes out the existing OpenMM examples directory to include three tutorial notebooks:

- openmm_polyalanine: a notebook that demonstrates how to set up an OpenMM simulation from scratch with NanoVer, use features of the OpenMM app in conjunction with NanoVer, and demonstrates how to record and playback simulations recorded directly using NanoVer
- openmm_nanotube: a notebook that demonstrates how to interact with an OpenMM NanoVer simulation from a python client
- openmm_neuraminidase: a notebook that demonstrates how to set up an OpenMM simulation from scratch and change the visualisations of different selections for a protein-ligand system.

All of these notebooks use the OmniRunner and OpenMMSimulation classes, and the polyalanine notebook demonstrates the recording and playback functionality of NanoVer (@Ragzouken could you check that I've done this correctly?). The recording and playback stuff may want to appear in a more general section at some point, but as the polyalanine notebook already demonstrated how to record trajectories using the OpenMM functionality it seemed like a good time to mention the NanoVer functionality.

Comments welcome!


examples/openmm/openmm_nanotube.ipynb
```diff
@@ -0,0 +1,594 @@
...
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# This will just automatically close previous instances of the runner if you re-run this cell multiple times\n",
+    "try:\n",
+    "    imd_runner.close()\n",
+    "except NameError:\n",
+    "    pass\n",
+    "imd_runner = OmniRunner.with_basic_server(nanotube_simulation, nanotube_simulation, name=\"nanotube-omm-server\")"
```
Ragzouken, line 171: You pass the simulation twice here

examples/openmm/nglclient.py
Ragzouken, line 1: Rather than duplicating this file here from the ase folder, maybe we should move it to the mdanalaysis module so it can be imported?

Ragzouken (approved):
Looks good! Omni usage, including recording, seems correct.

examples/openmm/nglclient.py
hjstroud: This is something Rhos and I have been discussing, it seems like at some point we should do a restructure to put any duplicated files in a more accessible location for all of the directories. Is it ok to leave this for another PR?

examples/openmm/openmm_nanotube.ipynb
hjstroud: Well spotted! I'll correct this now

hjstroud, line 171:
```suggestion
    "imd_runner = OmniRunner.with_basic_server(nanotube_simulation, name=\"nanotube-omm-server\")"
```

PR #220, opened 2024-08-12
Docstrings for OmniRunner etc

Ragzouken:
These docstrings may appear in the documentation website too, so it'd be good to get input on if they need work to fit in there.

### DIFF
--- a/python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
+++ b/python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
@@ -43,7 +43,7 @@ class ASEOpenMMSimulation:
         """
         Construct this from an existing ASE OpenMM simulation.
         :param simulation: An existing ASE OpenMM Simulation
-        :param name: An optional name for the simulation
+        :param name: An optional name for the simulation instead of default
         """
         sim = cls(name)
         sim.simulation = simulation
@@ -54,7 +54,7 @@ class ASEOpenMMSimulation:
         """
         Construct this from an existing NanoVer OpenMM XML file at a given path.
         :param path: Path of the NanoVer OpenMM XML file
-        :param name: An optional name for the simulation
+        :param name: An optional name for the simulation instead of filename
         """
         sim = cls(name or Path(path).stem)
         sim.xml_path = path
--- a/python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
+++ b/python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
@@ -34,14 +34,28 @@ class InitialState:
 
 
 class ASEOpenMMSimulation:
+    """
+    A wrapper for ASE OpenMM simulations so they can be run inside the OmniRunner.
+    """
+
     @classmethod
     def from_simulation(cls, simulation: Simulation, *, name: Optional[str] = None):
+        """
+        Construct this from an existing ASE OpenMM simulation.
+        :param simulation: An existing ASE OpenMM Simulation
+        :param name: An optional name for the simulation
+        """
         sim = cls(name)
         sim.simulation = simulation
         return sim
 
     @classmethod
     def from_xml_path(cls, path: PathLike[str], *, name: Optional[str] = None):
+        """
+        Construct this from an existing NanoVer OpenMM XML file at a given path.
+        :param path: Path of the NanoVer OpenMM XML file
+        :param name: An optional name for the simulation
+        """
         sim = cls(name or Path(path).stem)
         sim.xml_path = path
         return sim
@@ -69,6 +83,9 @@ class ASEOpenMMSimulation:
         self.checkpoint: Optional[InitialState] = None
 
     def load(self):
+        """
+        Load and set up the simulation if it isn't done already.
+        """
         if self.simulation is None and self.xml_path is not None:
             with open(self.xml_path) as infile:
                 self.simulation = serializer.deserialize_simulation(
@@ -93,6 +110,11 @@ class ASEOpenMMSimulation:
         )
 
     def reset(self, app_server: NanoverImdApplication):
+        """
+        Reset the simulation to its initial conditions, reset IMD interactions, and reset frames to begin with topology
+        and continue.
+        :param app_server: The app server hosting the frame publisher and imd state
+        """
         assert (
             self.simulation is not None
             and self.atoms is not None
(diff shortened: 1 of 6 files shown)


python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
```diff
@@ -34,14 +34,28 @@ class InitialState:
...
         sim = cls(name)
         sim.simulation = simulation
         return sim
 
     @classmethod
     def from_xml_path(cls, path: PathLike[str], *, name: Optional[str] = None):
+        """
+        Construct this from an existing NanoVer OpenMM XML file at a given path.
+        :param path: Path of the NanoVer OpenMM XML file
+        :param name: An optional name for the simulation
```
rhoslynroebuck, line 57: What is the default name for the simulation if none is given?

python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
```diff
@@ -93,6 +110,11 @@ def load(self):
         )
 
     def reset(self, app_server: NanoverImdApplication):
+        """
+        Reset the simulation to its initial conditions, reset IMD interactions, and reset frames to begin with topology
```
rhoslynroebuck, line 114:
```suggestion
        Reset the simulation to its initial conditions, reset IMD interactions, and reset frames to initial positions
```

rhoslynroebuck: Just a suggestion!

python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
```diff
@@ -149,12 +171,22 @@ def reset(self, app_server: NanoverImdApplication):
...
 
     def advance_by_one_step(self):
+        """
+        Advance the simulation to the next point a frame should be reported, and send that frame.
+        """
         self.advance_to_next_report()
 
     def advance_by_seconds(self, dt: float):
+        """
+        Advance the simulation playback by some seconds. This time is ignored and the next frame is generated and sent.
```
rhoslynroebuck, line 181: Not sure I fully understand, which time is ignored?

python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
```diff
@@ -149,12 +171,22 @@ def reset(self, app_server: NanoverImdApplication):
...
+        :param dt: Time to advance playback by
```
rhoslynroebuck, line 182:
```suggestion
        :param dt: Time to advance playback by in seconds
```

python-libraries/nanover-omni/src/nanover/omni/omni.py
```diff
@@ -99,40 +117,67 @@ def simulation(self):
...
+        :param index: Index of simulation to switch to
+        :return:
+        """
         self._cancel_run()
         self._simulation_index = int(index) % len(self.simulations)
         self._start_run()
 
     def next(self):
+        """
+        Switch to the next simulation in the list of available simulations.
```
rhoslynroebuck, line 146: What is the behaviour if we are at the end of the list of simulations?

python-libraries/nanover-omni/src/nanover/omni/omni.py
```diff
@@ -99,40 +117,67 @@ def simulation(self):
...
 
     def list(self):
+        """
+        Get the list of available simulations.
+        """
         return {"simulations": [simulation.name for simulation in self.simulations]}
 
     def reset(self):
+        """
+        Reset the currently active simulation to its initial state.
```
rhoslynroebuck, line 158:
```suggestion
        Reset the currently-active simulation to its initial state.
```

python-libraries/nanover-omni/src/nanover/omni/omni.py
```diff
@@ -99,40 +117,67 @@ def simulation(self):
...
+        """
         assert self._runner is not None
         self._runner.signals.put("reset")
 
     def pause(self):
+        """
+        Pause the currently active simulation.
```
rhoslynroebuck, line 165:
```suggestion
        Pause the currently-active simulation.
```

python-libraries/nanover-omni/src/nanover/omni/playback.py
```diff
@@ -17,6 +17,10 @@
 class PlaybackSimulation:
     @classmethod
     def from_paths(cls, paths: Iterable[PathLike[str]]):
+        """
+        Construct this from one or both of trajectory and state recording file path.
+        :param paths: One or both of trajectory and state recording file path.
```
rhoslynroebuck, line 22:
```suggestion
        :param paths: One or both of trajectory and state recording file paths.
```

python-libraries/nanover-omni/src/nanover/omni/playback.py
```diff
@@ -17,6 +17,10 @@
 class PlaybackSimulation:
     @classmethod
     def from_paths(cls, paths: Iterable[PathLike[str]]):
+        """
+        Construct this from one or both of trajectory and state recording file path.
```
rhoslynroebuck, line 21:
```suggestion
        Construct this from one or both of trajectory and state recording file paths.
```

rhoslynroebuck (commented):
Just left a few minor comments and suggestions

python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
Ragzouken: Have adjusted the wording, but here's an explanation: `advance_by_seconds` is used for all simulations because the recording playback should advance by a number of real world seconds so that it plays back 1:1 at the speed recorded. For live simulations that real world time doesn't correspond to the simulation time, and goes unused and the simulation is just advanced how many steps (`frame_interval`) until the next frame is generated.

python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
Ragzouken: Have clarified this last part--it refers to the fact that the sequence of topology->normal->normal->normal frames will begin again

python-libraries/nanover-omni/src/nanover/omni/ase_omm.py
Ragzouken: For a file it defaults to the filename, for an existing simulation it's something like `Unnamed ASE OpenMM Simulation" or something like that--I have updated these to be a little more specific

PR #223, opened 2024-08-13
Remove total energy field from FrameData for OpenMM

hjstroud:
This PR removes the "total" energy from the FrameData. The quantity that was previously being calculated was the sum of the kinetic and potential (system only, no user energy) energies. This is not a quantity that OpenMM calculates, and given that we deliver the kinetic, potential and user energies separately, it is probably more transparent to remove this field and allow a user to calculate the total energy themselves, at which point they can be clear about whether or not they are including the user energy in their calculation.


Ragzouken (approved):
Looks good!

PR #225, opened 2024-08-14
Clear state keys relevant to recording when looping playback

Ragzouken:


Ragzouken:
Also good idea to reset box pose to identity. Need to ignore locks when replaying state

hjstroud (approved):
Looks good! We've had a look at this in person and talked it through :)

PR #226, opened 2024-08-19
Add MDAnalysis/NanoVer notebook for analysis of NanoVer recordings

hjstroud:
This PR adds a notebook that details how to extract data from an iMD simulation recorded using NanoVer using MDAnalysis, and highlights some of the benefits of recording trajectories using the inbuilt NanoVer functionality.

As ever, comments and suggestions are welcome!


examples/mdanalysis/mdanalysis_nanover_recording.ipynb
rhoslynroebuck, line 1: Looks great! 

The only issue I had was that the first line in the notebook about matplotlib widgets requires the `ipympl` package to be installed, but I can't see this in the requirements in the docs, and it didn't get installed for me as a dependency for any of the other packages. Would be good to add a note to the documentation (possibly also in the notebook?) that this needs to be installed.

rhoslynroebuck (commented):
Looks great! See my comment on the notebook about dependencies

examples/mdanalysis/mdanalysis_nanover_recording.ipynb
hjstroud: Ah interesting, this was something I found to be automatically installed in my environment, but it's good to know that it's not there automatically. I'll add something, thanks!

rhoslynroebuck: Oh that's weird, maybe something about my env then? I'm running with jupyter notebook not jupyter lab, so could be something to do with that

rhoslynroebuck (approved):

hjstroud:
Whilst making this notebook I noticed that the units of the system and user forces are not being converted properly from the NanoVer units (kJ/(mol.nm)) to the MDAnalysis units (kJ/(mol.A)). I have flagged this in Issue #227, and this will be fixed with PR #228.

rhoslynroebuck (approved):
Fab, all good.

Issue #227, opened 2024-08-21
Units of forces are incorrect for NanoVer system and user forces when using MDAnalysis 


hjstroud:
When loading a NanoVer trajectory as an MDAnalysis universe, the units of the forces and user forces should be in units of kJ/(mol.Angstrom), but instead appear in units of kJ/(mol.nm) i.e. those used in NanoVer.

PR #228, opened 2024-08-21
Fix MDAnalysis force units when loading NanoVer trajectory as an MDAnalysis Universe

hjstroud:
This PR aims to fix Issue #227, converting the units of force used by NanoVer (i.e. kJ/(mol.nm)) to the units of force used by MDAnalysis (i.e. kJ/(mol.Angstrom)) for the user forces and system forces when loading a recorded NanoVer trajectory as an MDAnalysis universe.


rhoslynroebuck (approved):
Great. We've spoken through this in person and all looks good. One thing to note is that there are currently no tests to check any of the units (forces/velocities/positions), so this PR does not include a test. Might be worth writing some in the future.

PR #151, opened 2024-06-24
combined ase/omm/replay server

Ragzouken:

PR #231, opened 2024-08-23
Add NGLClient to NanoVer as a module

hjstroud:
This PR adds the existing NGLClient python script used in various places in the tutorials in `examples` to NanoVer as its own module. At this stage, all that we have done is add the existing file and replicate the structure of the other module directories in the new `nanover-nglview` directory, but there are a few things we're not sure about:

- Does anything need to be modified on the GitHub CI to make this work?
- Will adding this package with a new dependency interfere with existing users setups?
- Should this package be installed automatically with NanoVer or should it be optional (like LAMMPS), and how do we do this?
- Any other considerations?


Ragzouken:
You can see an example of the changes I made when adding the `nanover-omni` module here: https://github.com/IRL2/nanover-protocol/pull/151/files

I think we should avoid making the packages too granular, so perhaps renaming this to `nanover-notebooks` or `nanover-jupyter` or something would be better.

I don't see any problem with installing by default -- you do this by making a dependency of `nanover-server`

python-libraries/nanover-nglview/src/nanover/nglview/nglclient.py
```diff
@@ -0,0 +1,95 @@
...
+            self._view = frame_data_to_nglwidget(self.latest_frame)
+        return self._view
+
+    def _on_frame_received(self, frame_index: int, frame):
+        super()._on_frame_received(frame_index, frame)
+        self.view.set_coordinates(
+            {0: np.array(self.latest_frame.particle_positions) * 10}
+        )
+        if self.update_callback is not None:
+            self.update_callback(self.universe)
```
Ragzouken, line 34: The type checker doesn't like this because `universe` isn't defined on the class--and I don't see anywhere where it gets set either..

hjstroud: I noticed this... I can't tell if this was supposed to add functionality that was never actually completed or whether it does something I'm unaware of... Should we keep it?

hjstroud: For now, I have removed this if statement and replaced it with a TODO to implement functionality to allow the user to update callbacks used in the widget.

Ragzouken (approved):
Looks good to me!

Issue #230, opened 2024-08-21
ASEOpenMMSimulation/OmniRunner not working properly


hjstroud:
It is unclear exactly what the issue is, but there are a number of things that seem to be broken when running the ASE tutorial notebooks using the OmniRunner with an ASEOpenMMSimulation rather that using the old runners. The issues encountered so far include:

- All of the atoms appearing as hydrogens (white balls), regardless of how they are defined in the simulation input
- The topology of the systems don't seem to be being read correctly (when viewed using the NGLView client): the constituent atoms of the "graphene sheet" are not bonded in the graphene example, and in the nanotube example the methane atoms are not bonded but the nanotube appears to have many bonds
- Bonds seem to appear and disappear when the methane enters the nanotube (when using iMD-VR)
- The graphene example runs _really_ slowly, somewhere in the region of two fps

When using the old runners, none of these problems occur.

Ragzouken:
This should be the fix, if you could try it: https://github.com/IRL2/nanover-protocol/pull/232

PR #232, opened 2024-08-26
Use correct frame method for omni's ase omm

Ragzouken:


hjstroud (approved):
Looks great to me! I have tested this with the ASEOpenMMSimulation/OmniRunner version of the graphene and nanotube example notebooks, and this PR appears to fix the problems highlighted in Issue #230 (at least in terms of what I can see using the NGLView python client). Maybe these systems should be tested in VR too, but otherwise I'm happy for this to go through. Well done for fixing this so quickly!

Ragzouken:
Thanks for the review, I think all the issues visible in VR (bonding and atom coloring) are because the topology wasn't passing through correctly because of the issues here, so I'll just go ahead and merge

PR #234, opened 2024-08-27
Update README

rhoslynroebuck:
Update the README with the following changes:

* Updated `Tutorials` section - this was outdated, so have replaced this with the relevant links to the docs. 
* Separated the installation instructions for the user and developer. The user installation now points to the docs for the most up-to-date instructions.
* Separated the developer installation instructions into clearer sections based on operating system
* Fixed typos & updated wording where needed
* Added a Contents section at the top for easier navigation

Things I'm still not sure about:
* Do we keep the `Getting started` section at the beginning or do we just direct to the docs?
* Do we keep the [Running the examples](https://github.com/IRL2/nanover-protocol/tree/update-README?tab=readme-ov-file#running-the-examples) heading at the end of the developer installation section? They're not complete but might be useful?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,10 @@ For more information please take a look at [the project's documentation](https:/
 1. [Getting started](#Getting-started)
 2. [User installation](#User-installation)
 3. [Developer installation](#Developer-installation)
-4. [Troubleshooting](#Troubleshooting)
-5. [Citation and external libraries](#Citation-and-external-libraries)
+4. [Running the tests](#Running-the-tests)
+5. [Running the examples](#Running-the-examples)
+6. [Troubleshooting](#Troubleshooting)
+7. [Citation and external libraries](#Citation-and-external-libraries)
 
 ----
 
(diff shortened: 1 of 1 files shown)


rhoslynroebuck:
@Ragzouken, quick question. We currently state to update nanover using: `conda install nanover-server`. Should we change this to `conda update nanover-server` or is there a reason that we use the `install` instead?

Ragzouken (approved):
Looks good to me!

Ragzouken:
> Things I'm still not sure about:
> 
> * Do we keep the `Getting started` section at the beginning or do we just direct to the docs?
> * Do we keep the [Running the examples](https://github.com/IRL2/nanover-protocol/tree/update-README?tab=readme-ov-file#running-the-examples) heading at the end of the developer installation section? They're not complete but might be useful?

Maybe it's good to have something directly in the README even if the docs are better.

I think keeping the examples there is fine.

PR #237, opened 2024-08-29
Restructure fundamentals directory

hjstroud:
This PR adds a `basics` directory to the examples, and restructures the examples directory to place the `getting_started` and `nanover_nglview` notebooks in this directory (rather than in the `fundamentals` and `ase` directories, respectively). 

It also removes the now redundant `nglclient.py` file and updates the LAMMPS tutorial to use the `nanover.nglview` module instead.


rhoslynroebuck (approved):
All good to me!

Issue #239, opened 2024-08-29
Cannot attach DCD reporter without loading the simulation first (examples)


hjstroud:
In the `openmm_polyalanine.ipynb` example notebook, when running only the cells necessary to record a trajectory it is not possible to attach a DCD reporter to the simulation, and you instead get the following error: `AttributeError: 'NoneType' object has no attribute 'reporters'`. 

This can be fixed by loading the simulation _before_ attaching the reporter (i.e. `{sim_name}.load()`). In the short term at least, this notebook should be updated to reflect this. @Ragzouken is this expected behaviour?

rhoslynroebuck:
Closed by #240

PR #240, opened 2024-08-29
Fix DCD reporter section of example notebook

hjstroud:
As @ludovicaisa noticed, if you run only the functionality required to record a NanoVer trajectory using a DCD reporter as described by the `openmm_polyalanine` notebook, it is not possible to attach a DCD reporter to an OpenMMSimulation without first loading it using the `.load()` function. This PR adds a note to the notebook to specify this, and adds a line of code to do this. This addresses Issue #239.


rhoslynroebuck (approved):
Yeah, this is a good comment to add. Ran through this notebook and all good.

PR #242, opened 2024-09-19
Fix/update nanover diagrams in fundamentals

Ragzouken:


Ragzouken:
I've updated the diagrams to no longer be inaccurate and to use NanoVer not Narupa, however I'm not sure if the notebooks themselves are entirely useful. Would be useful to go over them together.

PR #244, opened 2024-09-25
Slightly improve docstrings for Omni

Ragzouken:


python-libraries/nanover-omni/src/nanover/omni/omni.py
```diff
@@ -107,17 +107,26 @@ def print_basic_info_and_wait(self):
...
+        """
         try:
             return self.simulations[self._simulation_index]
         except IndexError:
             return None
 
     @property
     def is_paused(self):
+        """
+        Is the currently selected simulation paused?
```
Mohamed-Dhouioui, line 128: """
    Checks if the currently selected simulation is paused.

    Returns:
        bool: True if the simulation is paused, False otherwise. Returns None if no runner is available.
    """

python-libraries/nanover-omni/src/nanover/omni/openmm.py
```diff
@@ -18,6 +18,9 @@
 class OpenMMSimulation:
     """
```
Mohamed-Dhouioui, line 19: """
    A wrapper for OpenMM simulations to run inside the OmniRunner.

    The following attributes can be configured after construction:
    - :attr:`frame_interval`: Number of simulation steps to advance between frames.
    - :attr:`include_velocities`: Include particle velocities in frames.
    - :attr:`include_forces`: Include particle forces in frames.
    - :attr:`platform_name`: Name of OpenMM platform to use when loading the system from XML.
    """

Mohamed-Dhouioui (changes requested):
I have commented on the lines that might need some edits. Other than that, Nicely done!

Mohamed-Dhouioui (approved):
This looks good. Nice job!

Issue #241, opened 2024-09-16
ASE simulations that don't use OpenMM don't have a simulation class suitable for the OmniRunner


hjstroud:
Currently we can run simulations using OpenMM with or without ASE (OpenMMSimulation and ASEOpenMMSimulation, respectively), but we don't have a corresponding ASESimulation class for running simulations using ASE via the OmniRunner. This is relevant for the `basic_example.ipynb` and `openmm_neuraminidase.ipynb` notebooks in the ase examples folder.

Ragzouken:
Looking over it, these are examples that deliberately avoid using the runners, instead using `NanoverASEDynamics`. I think it can be made to work with the Omni runners easily, but I think need to avoid reconstructing the dynamics when resetting.

PR #245, opened 2024-09-25
Non-omm ase

Ragzouken:
Fixes https://github.com/IRL2/nanover-protocol/issues/241

- [x] non-omm ase
- [x] update notebooks
- [x] add tests
- [x] docstrings


Ragzouken:
I've updated the `basic_example` notebook to do something functional, but it may need some restructuring to be sensible example

examples/ase/ase_basic_example.ipynb
```diff
@@ -430,6 +434,33 @@
   {
    "cell_type": "code",
    "execution_count": 17,
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "MissingDataError",
+     "evalue": "'\\'No value with the key \"particle.count\".\\''",
```
hjstroud, line 441: Currently there's no `particle_count` value field in the FrameData: I'm not sure if there are other keys that should be being populated that are currently not populated...

examples/ase/ase_openmm_neuraminidase.ipynb
```diff
@@ -342,59 +413,58 @@
...
+     "end_time": "2024-10-11T11:03:13.895040Z",
+     "start_time": "2024-10-11T11:03:13.889625Z"
+    }
+   },
    "outputs": [],
    "source": [
     "from ase.md import Langevin\n",
     "import ase.units as ase_units\n",
-    "dynamics = Langevin(atoms, timestep=1.0 * ase_units.fs, temperature_K=300 * ase_units.kB, friction=1.0e-03)"
+    "dynamics = Langevin(atoms, timestep=1.0 * ase_units.fs, temperature_K=300, friction=1.0e-03)"
```
hjstroud, line 433: When setting the temperature using `temperature_K` I think you only need to specify the number in Kelvin, rather than converting to units of kT by multiplying by the Boltzmann constant kB (see the first note on [this page](https://wiki.fysik.dtu.dk/ase/ase/md.html)

examples/ase/ase_basic_example.ipynb
hjstroud: Ignore this, this was because we were accessing `latest_frame` rather than `current_frame`: after changing this everything works just fine!

hjstroud (approved):
Have changed `latest_frame` to `current_frame` and everything seems to work just fine! Nice work!

Issue #154, opened 2024-06-26
`python-libraries/nanover-essd/tests/test_essd_client_server.py::test_remove_service` failing sometimes on windows (maybe others)


Ragzouken:

Ragzouken:
Hopefully fixed in https://github.com/IRL2/nanover-protocol/pull/246

Ragzouken:
Nope 😔

Ragzouken:
May be fixed in https://github.com/IRL2/nanover-protocol/pull/249

PR #250, opened 2024-10-01
Update developer install to modern package toml

Ragzouken:
re: https://github.com/pypa/pip/issues/11457

- `compile_proto` now a standalone script instead of a `setup.py` subcommand
- `setup.py` replaced with `pyproject.toml`
- compile scripts now look for `pyproject.toml` instead of `setup.py` to determine which folders are packages
- remove `bld.bat` in `nanover-core` conda package -- not clear why this is the only one with such a script


Ragzouken:
Much of the way there but the package installation doesn't work properly (in conda only?)

Ragzouken:
The compile scripts weren't correct, but now seem to be working, now only conda packaging tests are failing I think.

Ragzouken:
Just need to fix mypy CI now..

Mohamed-Dhouioui (approved):
I am not sure about the versioning for packages requirements here and the need for python>=3.12. However, since all the tests are running successfully, I am assuming that all is well. If issues arise in the future we can get back and check.

Mohamed-Dhouioui (commented):
I noticed that some tests are failing, The error occurs in the test `test_remove_service ` within the file `python-libraries/nanover-essd/tests/test_essd_client_server.py`. The assertion was checking that a ServiceHub instance was not in a set of ServiceHub instances, but it failed because it was found in the set.

@Ragzouken could you check before merging?

Ragzouken:
> I noticed that some tests are failing, The error occurs in the test `test_remove_service ` within the file `python-libraries/nanover-essd/tests/test_essd_client_server.py`. The assertion was checking that a ServiceHub instance was not in a set of ServiceHub instances, but it failed because it was found in the set.
> 
> @Ragzouken could you check before merging?

It's a recurring issue unrelated to this change.. it's irritatingly persistent https://github.com/IRL2/nanover-protocol/issues/154

Issue #248, opened 2024-09-30
"Protobuf gencode version 5.27.2 is older than the runtime version 5.28.2" (Mac OS)


phuddha:
Gpoing through a clean developer installation on Mac OS - following readme. 

On running tests ( python -m pytest python-libraries) there is 1 failure and 6 warnings:

FAILED python-libraries/nanover-essd/tests/test_essd_client_server.py::test_remove_service - AssertionError: assert ServiceHub(**{'name': 'test service', 'address': '192.168.0.7', 'services': {'trajectory': 54321, 'imd': 54322, 'multiplayer': 54323, 'builder': 54324...

--- warnings ---

../../../../../opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112
  /opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112: UserWarning: Protobuf gencode version 5.27.2 is older than the runtime version 5.28.2 at nanover/protocol/command/command_service.proto. Please avoid checked-in Protobuf gencode that can be obsolete.
    warnings.warn(

../../../../../opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112
  /opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112: UserWarning: Protobuf gencode version 5.27.2 is older than the runtime version 5.28.2 at nanover/protocol/state/state_service.proto. Please avoid checked-in Protobuf gencode that can be obsolete.
    warnings.warn(

../../../../../opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112
  /opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112: UserWarning: Protobuf gencode version 5.27.2 is older than the runtime version 5.28.2 at nanover/protocol/trajectory/get_frame.proto. Please avoid checked-in Protobuf gencode that can be obsolete.
    warnings.warn(

../../../../../opt/anaconda3/envs/nanover-dev/lib/python3.12/site-packages/google/protobuf/runtime_version.py:112
[...]

Issue #43, opened 2024-02-26
Compilation issue with grpcio-tools 1.62


jbarnoud:
This is the error:

```
Traceback (most recent call last):
  File "/home/runner/work/narupa-protocol/narupa-protocol/./python-libraries/nanover-core/setup.py", line 146, in <module>
    setup(
  File "/usr/share/miniconda3/envs/test/lib/python3.12/site-packages/setuptools/_distutils/core.py", line 185, in setup
    return run_commands(dist)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/test/lib/python3.12/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
    dist.run_commands()
  File "/usr/share/miniconda3/envs/test/lib/python3.12/site-packages/setuptools/_distutils/dist.py", line 969, in run_commands
    self.run_command(cmd)
  File "/usr/share/miniconda3/envs/test/lib/python3.12/site-packages/setuptools/dist.py", line 963, in run_command
    super().run_command(command)
  File "/usr/share/miniconda3/envs/test/lib/python3.12/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
    cmd_obj.run()
  File "/home/runner/work/narupa-protocol/narupa-protocol/./python-libraries/nanover-core/setup.py", line 62, in run
    compile_protocol(self.proto_dir, setup_path / "src", self)
  File "/home/runner/work/narupa-protocol/narupa-protocol/./python-libraries/nanover-core/setup.py", line 82, in compile_protocol
    proto_include = protoc.pkg_resources.resource_filename("grpc_tools", "_proto")
                    ^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'grpc_tools.protoc' has no attribute 'pkg_resources'. Did you mean: 'resources'?
```

In commit b52cb8f2a1d3aeba26038e031c38211039a42900, I pin `grpcio-tools` to a version less than 1.62.

PR #252, opened 2024-10-07
Don't use old grpcio-tools

Ragzouken:
Might fix https://github.com/IRL2/nanover-protocol/issues/248


Ragzouken:
Was pinned lower in https://github.com/IRL2/nanover-protocol/issues/43 but seems to work now

PR #258, opened 2024-10-10
Adding non-expert input files from "Flexible Docking" paper

ludovicaisa:
This pull request adds input files (.xml) for non expert users for the systems examined in [this paper](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0228461) .
These files were obtained converting old NarupaXR files to Nanover iMD-VR format.

### DIFF
--- a/examples/basics/openmm_files/system_descriptions.md
+++ b/examples/basics/openmm_files/system_descriptions.md
@@ -1,0 +1,16 @@
+# Flexible Protein-Ligand Docking Paper's files
+
+This repository contains input simulations files from 
+["Interactive molecular dynamics in virtual  reality for accurate flexible protein-ligand docking"](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0228461).
+which are:
+
+1. ** 2g5n_complex.xml ** : trypsin and indole-amidine, based on PDB file 2G5N. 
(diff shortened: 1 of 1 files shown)


hjstroud (commented):
It may be worth renaming the files to make it very clear what systems the files contain, e.g. if the file contains a neuraminidase and oseltamivir complex, it could be called neuraminidase_oseltamivir_complex.xml

Would it also be worth adding a text file (e.g. file_descriptions.md) that contains a description of what each of the systems are, and perhaps a reference to the paper(s) that they come from? @Ragzouken @rhoslynroebuck this may be something worth adding to any directory with example system files in, but adding these files to other directories is outside the scope of this PR.

ludovicaisa:
> the file contains a neuraminidase and oseltamivir complex, it could be called neuraminidase_oseltamivir_complex.xml

The names of the files contain their PDB reference code as they appear in the paper. It would be good to leave them as they are for this reason.

hjstroud (approved):
Great work adding a descriptive file, this is something we should make standard for any directory containing example input files. And it makes sense to leave the naming conventions to match the paper. Nice work, hit that Squash & Merge button! 😄

hjstroud (approved):

PR #260, opened 2024-10-10
Revert "Don't use old grpcio-tools"

Ragzouken:
Reverts IRL2/nanover-protocol#252

Seems to cause issues that aren't caught by the tests


Ragzouken:
https://github.com/IRL2/nanover-protocol/issues/261

PR #262, opened 2024-10-10
Specify conda dependencies at environment creation

Ragzouken:
This way conda can pick the appropriate python version e.g openmm doesn't necessarily support the latest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,8 @@ page in our documentation for detailed instructions on installing NanoVer.
 * Install Anaconda
 * Clone the nanover-protocol repository
 * In a terminal, in the repository root:
-    * Create a conda environment (here we call the environment "nanover-dev"): `conda create -n nanover-dev "python>3.11"`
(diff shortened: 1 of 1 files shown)


rhoslynroebuck:
Just to be explicit, this will mean that our installation instructions will have to be rewritten such that the user must conda install `nanover-server` before `nanover-imd` (currently the instructions are modular so that the user can install either package independently). I don't think this is a problem, but I know you had a couple of concerns @hjstroud

hjstroud (approved):
Looks good to me! Have changed the docs to reflect this new method of environment creation and submitted a PR for it.

hjstroud:
> Just to be explicit, this will mean that our installation instructions will have to be rewritten such that the user must conda install `nanover-server` before `nanover-imd` (currently the instructions are modular so that the user can install either package independently). I don't think this is a problem, but I know you had a couple of concerns @hjstroud

I did, but I agree with you that it's not a problem.

PR #265, opened 2024-10-10
Provide system temperature

Ragzouken:


Ragzouken:
Best wait for https://github.com/IRL2/nanover-protocol/pull/245 which changes the ase simulations around

hjstroud:
Need to make sure that we calculate the _instantaneous temperature of the system_ (NOT the temperature of the thermostat) for the programs we support:

- [ ] Calculate instantaneous temperature in OpenMM via kinetic energy for runners without inbuilt methods
- [ ] Calculate instantaneous temperature in OpenMM via inbuilt methods for runners with this functionality (perhaps we don't want to support this?)
- [ ] Calculate instantaneous temperature in ASE (i.e. cannot just take the temperature of the thermostat)

Ragzouken:
First draft for OpenMM based on existing StateDataReporter code -- untested

Ragzouken:
We should also add this field for MDAnalysis converter

hjstroud:
> We should also add this field for MDAnalysis converter

After testing, I think this data field can already be retrieved via the converter, so no need!

hjstroud:
Added tests for both OpenMM and ASE implementations:

- OpenMM implementation checks that the correct temperature is delivered when there are no iMD interactions, and that the temperature delivered is within 1% of the true temperature when there are iMD interactions
- ASE implementation checks that the temperature delivered in the frame matches that calculated internally by ASE

PR #267, opened 2024-10-11
Added a Python GUI

Mohamed-Dhouioui:
This pull request introduces a python GUI replication the Rust one to the nanover-protocol repository:

### Changes
Add a New Python GUI:

- Implemented a new user interface using Gradio. Please install the library to use I by running `pip install gradio`

- Created functions to run and stop simulations with various configurations.

- Added a detailed Jupyter Notebook (examples/basics/using_GUI.ipynb) explaining how to use the new GUI.

### Features _implemented_:

- [x] Running from Xml Files
- [x] Running from recordings
- [x] Specify Server Name
- [x] Specify Server Port
- [x] Specify Frame interval
- [x] check box to include velocities
- [x] check box to include frames
- [x] Record trajectory and shared state (both should be selected or none, for the moment we can't record just one)
- [x] Check box to start simulation paused

### Features _missing_:

- [x] Simulation FPS
- [x] Force Interval
- [x] Record statistics
- [x] Statistics FPS


Ragzouken:
Perhaps the notebook ought to be renamed to "server ui" or "runner ui" or something more descriptive?

Ragzouken:
As discussed offline, we can drop these options + ui elements:
- Force Interval
- Record statistics
- Statistics FPS

We don't see a usecase for force interval independent of frame interval, and it complicates the implementation for work done etc. We don't use the statistics from rust, but can reintroduce whatever we may need later when debugging etc.

Ragzouken:
I've hooked up the simulation fps option

Ragzouken (changes requested):
Looks good to me, except that you should remove the features that we won't support, and maybe name the notebook for descriptively. Functionally it matches the rust server well, as far as I can tell 👍

examples/basics/nanover_nglview.ipynb
hjstroud, line 1: I don't think the changes that have been made to this file were supposed to be made in this PR. Could you undo these changes?

examples/basics/UI.py
```diff
@@ -0,0 +1,246 @@
...
+                    ["Normal", "Verbose", "Super verbose"],
+                    label="Verbosity Level",
+                    value="Normal",
+                )
+                show_progression = gr.Checkbox(label="Show simulation progression")
+
+                gr.Markdown("## Network")
+                # Textbox for server name
+                server_name = gr.Textbox(
+                    label="Server name", value="NanoVer-RS iMD Server"
```
hjstroud, line 156: The name "NanoVer-RS" refers to the Rust server. This should be changed, perhaps to simply "NanoVer"? This also applies to the automatic file names for the output trajectory and state files.

Another thing is that if this name is being used to name files by default (does the rust UI do this?), we may wish to swap out the spaces for either underscores or hyphens, so that the names are easier to type in the command line!

examples/basics/UI.py
```diff
@@ -0,0 +1,246 @@
...
+                with gr.Group():
+                    # Checkboxes and textboxes for recording options
+                    record_stats = gr.Checkbox(label="Record statistics")
+                    stats_file = gr.Textbox(label="Statistics file")
+                    stats_fps = gr.Number(label="Statistics FPS", value=4)
+                with gr.Group():
+                    record_trajectory = gr.Checkbox(label="Record trajectory")
+                    trajectory_output_file = gr.Textbox(
+                        label="Trajectory output file",
+                        value=f"{server_name.value}.traj",
```
hjstroud, line 185: See comments on server name

examples/basics/UI.py
```diff
@@ -0,0 +1,246 @@
...
+                    )
+                with gr.Group():
+                    record_shared_state = gr.Checkbox(label="Record shared state")
+                    shared_state_file = gr.Textbox(
+                        label="Shared state file", value=f"{server_name.value}.state"
```
hjstroud, line 190: See comments on server name

examples/basics/UI.py
hjstroud: Do we want to automatically provide a default name for the output files? In the rust server there is no default name or path for the input files, so if we're looking for parity with this we may wish to remove the default names inserted here

hjstroud (changes requested):
In addition to the single comments I have left in the files, I have a few suggestions:

- There is not yet a button to return the network parameters to their default values (apart from the FPS, which has an "undo"-type button. It may be worth adding buttons to restore the default values, labelling clearly what they are
- The UI looks nice, but in the Jupyter interface I can only see a small window of it at a time! Would it be possible to tell the cell to output it "full size", so a user can see all of the options at once?
- This may be beyond the scope of this PR, but the Rust GUI also has the option to "Run demonstration input!": @Ragzouken it may be worth us considering whether there's a system that we can include as a demonstration input for the user to try out the UI without having to load their own system?
- The Jupyter notebook nicely introduces the UI, but it doesn't give the user any guidance as to how to use it: it may be worth developing the notebook further so that it instructs the user on how to use the UI, perhaps with some example input file/s (there should already be some in this directory) to turn it into more of a tutorial-style notebook.

examples/basics/using_GUI.ipynb
```diff
@@ -0,0 +1,161 @@
+{
+ "cells": [
+  {
+   "metadata": {},
+   "cell_type": "markdown",
+   "source": "# Getting Started with usign the Nanover Python GUI",
```
hjstroud, line 6:
```suggestion
   "source": "# Getting Started with using the NanoVer Python GUI",
```
All references to NanoVer should capitalise the N and the V

examples/basics/UI.py
```diff
@@ -0,0 +1,230 @@
...
+):
+    """
+    Runs a simulation with the given parameters.
+
+    :param simulation_type: choose between "From xml" and "From recording"
+    :param input_files: list of input files to run as live simulation
+    :param trajectory_files: list of trajectory files to run as playback simulation
+    :param state_file: list of state files to run as playback simulation
+    :param verbosity: choose between "Normal", "Verbose", "Super verbose"
+    :param show_progression: show simulation progression
```
Ragzouken, line 32: This doesn't seem to do anything currently -- if it's easy to have it match rust (periodically print out current frame etc) then may as well do it, otherwise I don't think it's very important right now?

examples/basics/UI.py
```diff
@@ -0,0 +1,230 @@
...
+    shared_state_file,
+):
+    """
+    Runs a simulation with the given parameters.
+
+    :param simulation_type: choose between "From xml" and "From recording"
+    :param input_files: list of input files to run as live simulation
+    :param trajectory_files: list of trajectory files to run as playback simulation
+    :param state_file: list of state files to run as playback simulation
+    :param verbosity: choose between "Normal", "Verbose", "Super verbose"
```
Ragzouken, line 31: I think this also doesn't do anything currently? I think it's only for debug information regarding the server setup, so we could omit it for now?

Mohamed-Dhouioui: I've removed this for now.

examples/basics/UI.py
Mohamed-Dhouioui: I've removed this for now.

Ragzouken (approved):
Looks good to me!

examples/basics/runner_GUI.ipynb
```diff
@@ -0,0 +1,220 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "%%blocks\n",
+    "demo = UI.create_ui()\n",
+    "demo"
```
hjstroud, line 164: When I run this cell, the UI is still only partially visible (you have to scroll through the output of the cell, you cannot view it all at once in the notebook). Is this something you can address, i.e. make it so that the whole UI is visible in the output of the cell? It seems more difficult to use the UI when you can't see all of the parameters simultaneously

hjstroud (dismissed):
We chatted about these in person, but just to reiterate the things I noticed in reviewing the UI:

- It may be worth linking to the Gradio website when you mention the package (see comment in code)
- Currently only a fraction of the UI is only displayed in the notebook format: it may be worth enlarging the output box if possible
- When I tried loading a .traj file, the simulation wouldn't start and an error was thrown to say that I wasn't loading a file with the extension `.traj` (even though I was) - not sure what error is happening here but it would be good to address so we can do playback
- Can the UI support multiple simulation files/trajectory files being uploaded to run multiple simulations? I couldn't tell, but it seemed as though when I tried to switch between "From xml" and "From recording", only one of them could be selected at a time.  This is a feature that the rust UI supports, but I don't know if we would want to implement this later (i.e. in a separate PR).

I ran a simulation via the UI, and although it was a little slow to start it worked fine, so I think this is nearly there! Hopefully just some final tweaks remain to get everything working 🙂

hjstroud:
FYI, the .traj and .state files I used that didn't work were ones generated via the gui for a previous simulation that I recorded: maybe it could be a problem with the way the GUI writes the files?

examples/basics/UI.py
```diff
@@ -0,0 +1,217 @@
+import os.path
+
+import gradio as gr
```
Ragzouken, line 3: Since gradio is not added as a dependency, it might be helpful to add something like the following:

```suggestion
try:
    import gradio as gr
except ModuleNotFoundError:
    print(
        "Unable to import the gradio module. Try `pip install gradio`."
    )
    sys.exit(1)
```

hjstroud:
Have just tested this, and I could load multiple recordings at once (trajectories and states) as well as a simulation xml, but I noticed some weird behaviour when I did this:
- When I connected to the server, the first simulation was always a recording, no matter which order I defined the files in: is this expected behaviour?
- When I loaded the files with the "From Recording" box checked, the following was printed to the terminal:
  
  `Simulation started with type: settings: {'simulation_type': 'From recording', 'input_files': {...`
  
  where there is only one "simulation_type" (i.e. from recording) but two distinct types of simulation (one xml and two recordings). Also, when I did this and then tried to load the simulation xml from within NanoVer iMD, there was a force (not defined in the input file) being applied to the system that looked the same as the force applied in the recording: could these somehow be being confused as it thinks it's supposed to be a recording?
- When I repeated the procedure but loaded the simulation with the "From xml" box checked, I got an analogous message:

  `Simulation started with type: settings: {'simulation_type': 'From xml', 'input_files': {...`

  where again there is only one "simulation_type" (i.e. from xml) but two distinct types of simulation (one xml and two recordings). This time the simulation file behaved as normal (although it was still a recording that seemed to be loaded first (or possibly the xml with a fictitious force, but when I went back onto the xml simulation everything seemed fine.
  

We've talked about these in person. If this can be fixed (and tested to make sure it works!) then I think I'm happy with how the GUI is.

Ragzouken:
This issue may be relevant too: https://github.com/IRL2/nanover-protocol/issues/303

Mohamed-Dhouioui:
With this commit [378c6b8](https://github.com/IRL2/nanover-protocol/pull/267/commits/378c6b8252880948923dd0f3b89ab7532bb6da79) I have fixed the output message and replaced the printing of simulation type with a list of currently loaded files.
However, for the interaction forces that are not cleaned when changing between a recording playback and a simulation, I've tested this outside the GUI with the `OmniRunner.wit_basic_server()` module and it showed the same behavior. So, I believe this is a bug with `OmniRunner` not the GUI. I have created an issue for #302 and detailed the steps to recreate it.

Ragzouken:
Issue with the interactions/avatar etc persisting should be fixed on main.

Mohamed-Dhouioui:
@hjstroud I think this is ready to be merged now. Could you please check to see if everything works as expected on Mac?

Ragzouken:
> @hjstroud I think this is ready to be merged now. Could you please check to see if everything works as expected on Mac?

I spoke with Harry just now to catch him up on the dev meeting and he said he was satisfied with the changes. I've looked also and everything seems to be working as expected. I think you can go ahead and merge 👍

Issue #198, opened 2024-07-24
Go through ASE notebooks


hjstroud:
These notebooks provide an overview of the available features of the NanoVer ASE interface, so this should only be looking to make sure they read clearly and that things are explained properly (and that the notebooks run).

- [x] `basic_example.ipynb`
- [x] `openmm_graphene.ipynb`
- [x] `openmm_nanotube.ipynb`
- [x] `openmm_neuraminidase.ipynb`

The NGLView notebook is not included here: see Issue [59](https://github.com/IRL2/nanover-docs/issues/59) in the nanover-docs repo.

hjstroud:
These notebooks have all been renamed and updated. The only outstanding problems that I can see in these notebooks are addressed in Issue #273 the comment in PR #274. I think that both of these issues are beyond the scope of the present issue, so I will close this issue for now.

PR #274, opened 2024-10-23
Update the ASE notebooks that use the `ASESimulation` class

hjstroud:
This PR aims to complete the remaining tasks in Issue #198, updating the notebooks to contain current information about how NanoVer works, use the correct simulation and runner classes, correct mistakes in the notebooks and improve the information present in the notebooks. 

One thing that has emerged from doing this is that it is unclear how the setup for the neuraminidase simulation in `ase_openmm_neuraminidase.ipynb` (which uses the ASESimulation) differs from the setup of an ASEOpenMMSimulation (as in the `ase_openmm_graphene.ipynb` notebook) apart from the fact that in the former the integrator parameters are explicitly set, which is not yet possible in the latter. 

I may be misunderstanding exactly what's going on, @Ragzouken is this correct? If so, it seems like it would be sensible to be able to set the simulation parameters in the ASEOpenMMSimulation, as I can't see a case where someone would want to run a simulation without defining the parameters they are using. This would mean we could either:

1. Move to using the ASEOpenMMSimulation class for the `ase_openmm_neuraminidase.ipynb`
2. Explain in the `ase_openmm_neuraminidase.ipynb` that the setup demonstrated there is equivalent to using ASEOpenMMSimulation (which it currently isn't)

### DIFF
--- a/examples/ase/ase_basic_example.ipynb
+++ b/examples/ase/ase_basic_example.ipynb
@@ -11,8 +11,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This notebook walks through setting up a simulation with ASE and combining it with NanoVer to make an interactive molecular dynamics simulation. \n",
-    "We then connect to the simulation from a client, so we can look at the data being produced and visualize it."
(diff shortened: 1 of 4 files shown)


rhoslynroebuck (approved):
Yeah the tutorials work for me and they look good.

PR #276, opened 2024-10-28
Examples/adding newsims

ludovicaisa:
Changed not functioning 6hcx simulation file (.xml) with a new one of the same system, equally parametrized. 
Loaded notebook multiple_simulations.ipynb, with visuals.py for rendering  (in a directory called visualisation_demos) and MultipleFiles.gif to show how to change system inside Nanover.


examples/basics/multiple_simulations.ipynb
```diff
@@ -0,0 +1,514 @@
...
+    "from examples.basics.visualisations_demos.visuals import clear_selections"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3d39b840d35707e4",
+   "metadata": {},
+   "source": [
+    "Then let's load the simulation files: we are loading three different protein ligand complexes:\n",
+    "1. Neuraminidase and zanamavir\n",
```
hjstroud, line 54:
```suggestion
    "1. Neuraminidase and zanamivir\n",
```
Check that the spelling is correct throughout the notebook

hjstroud (changes requested):
This is shaping up nicely, good start!

I tried the notebook out myself and ran into a few problems:

1. The neuraminidase-zanamivir system appears outside of the simulation box, but the other systems appear inside the simulation boxes. This would be worth fixing!
2. When I tried to render the systems the proteins appeared as white strings with atoms dotted along the backbone: is this what you saw when you tested the notebook? (The rendering of the ligands is fine though!)
3. It would be worth adding a sentence about closing the server in the `Close the server` section (see other notebooks for examples of this)
4. At the moment the notebook doesn't contain a "Next steps" section: This is something that we have generally included in all of the tutorial notebooks, and would be good to put in here. Again, check out the other notebooks for examples of this.

There is also a PR in progress at the moment that deals with selections and switching selections (see #205): once this PR goes through it would be good to update the way that this notebook deals with setting the renderings of the simulations to follow that new procedure, and add description about this feature within this notebook.

This is a great start though, well done for getting the neuraminidase system working, I know that was causing some issues!

examples/basics/multiple_simulations.ipynb
hjstroud: Also, which input file are you using for this system? If you are not already doing so, it would probably be good to use the one adapted (and subsequently fixed) from the paper that the other input files have come from, rather than the version you made yourself (for consistency, and also in accordance with the information in the system_descriptions.md file)

hjstroud:
Further to this: #205 has now been merged into main, so we should update this branch and see if we can switch between selections this way!

examples/basics/multiple_simulations.ipynb
ludovicaisa: Actually the file is the Neuraminidase zanamivir file I have produced, but I didn't figure out how to initialize the protein inside the Nanover box 
( p.s.: it's just a visualization effect since implicit solvent simulation do not have boundary condition. )
But since we have the old file working I'll switch them.

ludovicaisa:
I sent pull request after the following changes due to the comment of the reviewer, I respond to your comments @hjstroud sequentially.
1. The final .xml file of neuraminidase-zanamivir complex is the one from the paper, fixed. It's centered in the box during the simulation.
2. The rendering of the proteins has been changed to `geometric spline` since `cartoon` rendering may shows white string from time to time; from what I understood this issue is already known. I may suggest to open an possible issue if there is not one yet.
3-4. comments on the Close Server section and new Next steps section have been added, give me a feedback about them.
Thank you for the your feedback!

hjstroud:
The notebook is looking awesome, well done!

I had a look at this today on a standalone quest, and I found a couple of problems (which may be related to the code rather than the notebook, maybe @Ragzouken can clarify?):

- When you run the first cell that contains ".load(0)" and ".play()", the simulation doesn't seem to be loaded: it looks like an empty box, there are no molecules in there (at least in VR, I haven't tried connecting from a different client)
- It seems that switching between the simulations using ".next()" in the Jupyter interface is fine, but when trying to switch between them in the VR interface sometimes the ligand isn't loaded...is this a problem when running using a different setup (e.g. quest link)?

@ludovicaisa Did you test this out in VR on your PC, and if so what setup did you use?

ludovicaisa:
> y seeing the issues described when trying on Quest standalone, but it's through no fault of the notebook that I can see. I will have to look into it further, but there may be one or both of a fault in the selection switching logic of omni server (perhaps timing issue) and the issues with the cartoon rendering breaking other stuff down



> The notebook is looking awesome, well done!
> 
> I had a look at this today on a standalone quest, and I found a couple of problems (which may be related to the code rather than the notebook, maybe @Ragzouken can clarify?):
> 
> * When you run the first cell that contains ".load(0)" and ".play()", the simulation doesn't seem to be loaded: it looks like an empty box, there are no molecules in there (at least in VR, I haven't tried connecting from a different client)
> * It seems that switching between the simulations using ".next()" in the Jupyter interface is fine, but when trying to switch between them in the VR interface sometimes the ligand isn't loaded...is this a problem when running using a different setup (e.g. quest link)?
> 
> @ludovicaisa Did you test this out in VR on your PC, and if so what setup did you use?

@hjstroud @Ragzouken I used the PC-VR tethered setup.

ludovicaisa:
Me and @Mohamed-Dhouioui were playing with the systems using the new APK (passthrough). 
I confirm that sometimes the ligand is absent but selecting the simulation file another time it appears again.
If one player changes the simulation file it works for the others too but there can be issues with the interaction that may not work.

Ragzouken:
Todo: Update from main and see what issues still there and we can move on with figuring those out

ludovicaisa:
I tested this pull [request](https://github.com/IRL2/nanover-server-py/pull/323#pullrequestreview-2472577675) on the notebook,  cartoon rendering no longer shows the previous problem.
The only flaw I see is that the ligand is not always displayed when you go from one simulation to another, but double-clicking on the sim file will bring it back.

Ragzouken:
The missing ligands etc should be fixed in this release of the VR client: https://github.com/IRL2/nanover-imd-vr/releases/tag/2024.12.03

ludovicaisa:
I just uploaded the notebook, after checking the ligand issuee with [this](https://github.com/IRL2/nanover-imd-vr/releases/tag/2024.12.03) VR client, it works fine for both PC-VR and APK.
I wait @hjstroud for checking and approving the pull request.

Ragzouken:
Have spoken to Harry and he's happy for us to merge after I verify for myself that it works as intended now. I've done that, so we can proceed!

Issue #176, opened 2024-07-09
Deprecate runners in favour of nanover-omni


Ragzouken:
Each runner uses its own version of the threading, networking, and command boilerplate which is difficult to maintain.

PR #278, opened 2024-10-29
Remove old ASEOpenMMRunner

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/176


hjstroud (approved):
This looks good to me. I've made a minor change in adding the `.pause()` command when starting the simulation to match the comment, and renaming the Narupa object to a NanoVer object. I haven't checked how all of these visualisations look in VR, but if they look fine there then I'm happy for this to go ahead!

Issue #208, opened 2024-08-07
Cannot specify simulation parameters with ASE-OMM CLI


rhoslynroebuck:
The `ase-omm` CLI command takes (requires?) as an input a NanoVer-OpenMM simulation xml, however, this runner does not actually read the `Integrator` field from this xml. This means that you can't specify dynamics parameters (e.g. temperature, timestep) from the command line and these parameters are actually set to some default values. It seems misleading to provide this CLI functionality if these parameters are ignored, and can't actually be changed from the default parameters using this interface (this can only be done from a python script).

It seems like the only use case would be if (a) we provide a way of writing a NanoVer-ASE-OMM xml that specifies parameters that are read the by the `ase-omm` runner AND (b) these parameters differ somehow from the available OpenMM parameters, otherwise you could just use the standard OpenMM runner. 

Maybe we should consider deprecating this CLI for the moment and think about if it's worth doing the above.

If any of this information is incorrect or anything vital is missing, please do say!

Ragzouken:
In a recent merge both `nanover-ase-omm` and `nanover-omm-server` were replaced to just tell you to use `nanover-omni --omm` and `nanover-omni --ase-omm`. There's no CLI arguments to do anything with the integrator here either though. I don't think the runners even offer any explicit options for this other than letting you access the integrator yourself to change values.

I agree that it probably doesn't make sense to support ase by CLI since it doesn't fully follow the input XMLs and the main motivation for ASE is having deeper access in python to dynamics (which you obviously don't have on the CLI). Only other thing I can think of is that ASE supports the boundary walls and regular OMM doesn't? Is that right?

The generic CLI even for `nanover-omni` will always be limited and maybe it's only really fit for demos vs. say launching it from a bespoke python script or jupyter notebook where you can set up exactly what you want. Maybe it would be good for us to figure out exactly what use cases for CLI we want to support vs what ought to be done in custom notebooks or setup scripts?

PR #279, opened 2024-10-30
Remove ASE OMM command line

Ragzouken:
XML files are valid OMM systems and you don't have any of the benefits of using ASE in the CLI. The ASE OMM CLI also ignores the integrator settings which is confusing. Either use OMM CLI or set up ASE explicitly.

Closes https://github.com/IRL2/nanover-protocol/issues/208


hjstroud (approved):
Yeah looks good to me. I think removing this option makes sense, at least until we establish what to do about using OpenMM XML files for ASEOpenMMSimulations.

PR #281, opened 2024-10-31
Clean up ASE OMM

Ragzouken:
Removing obsolete methods for settings up ASE dynamics, the associated tests, obsolete READMEs that have been missed in docs updates (and therefore probably not discoverable in the first place), removing ASE OMM loading from XML which isn't well defined, and stripping back the ASE OMM simulation type towards maybe removing it later.


examples/ase/ase_openmm_nanotube.ipynb
```diff
@@ -428,20 +434,52 @@
...
   {
    "cell_type": "code",
+   "execution_count": 12,
    "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "C:\\Users\\ragzo\\anaconda3\\envs\\nanover-dev\\Lib\\site-packages\\MDAnalysis\\coordinates\\PDB.py:777: UserWarning: Unit cell dimensions not found. CRYST1 record set to unitary values.\n",
```
hjstroud, line 451: We might want to remove the output of this cell.

python-libraries/nanover-ase/src/nanover/ase/frame_adaptor.py
hjstroud, line 1: This file seems like an important one: has `frame_adapter` been superseded by something else, or is it still used somewhere?

Ragzouken: This file should not be referenced anywhere anymore. It basically encapsulated the idea of sending the topology first then sending normal frames, and tracking the frame index. This now instead happens in the omni simulations rather than baking it into each type of frame extraction method

hjstroud (changes requested):
We've been through all of this thoroughly in person, and I think it all makes sense. We should be extra careful to check that nothing breaks as a result of this, but I think you've gone through and checked the notebooks and we've put an issue up about the need to test things that are not currently tested, which can be dealt with separately. I have left a comment to add docstrings for any new functions added in this PR, but as long as that is done I'm fine with this going through. Good job!

hjstroud (approved):
Awesome, nice work!

Issue #273, opened 2024-10-22
Simulation in `ase_basic_example.ipynb` is running very slowly


hjstroud:
When connecting a VR client to the copper matrix simulation in the ASE basic example, the simulation appears to be running with a very low frame rate. Not sure if this is a result of the new runner, I remember this being an issue a couple of months ago, maybe for a different system: @Ragzouken do you know what's going on here?

Ragzouken:
Will have to look into it, but there was an issue before that ASE was guessing the bonds every frame instead of just for the first frame--if it's that it should be an easy fix.

Ragzouken:
Don't think it's that. Advance to next report takes 160ms apparently, which seems a lot..

Ragzouken:
Actually, it seems all time is spent running 5 frames of ASE dynamics, so I think ASE is just that slow, at least for that system. We could reduce the frame interval to make it smoother (but advance less simulation time per real world second)

hjstroud:
I see. It may be that the specific calculator used in the example takes a long time to calculate each step, but if we have a lot of examples that use the OMM calculator (which seems fast enough), perhaps the best course of action would be to explicitly set the frame rate in the notebook itself for the specific example and add a comment to explain what's happening, but leave the default settings on the server as 5 simulation steps per frame. I will do this and check whether this fixes the behaviour.

PR #282, opened 2024-10-31
Change frame interval and next() to load(0) in ase basic notebook

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/273


hjstroud (approved):
This looks great. I've tested this in VR and everything seems to be working much better, the dynamics are smooth again. Good job spotting this fix!

Issue #283, opened 2024-10-31
Missing ability to save/load OpenMM serialised state


hjstroud:
Currently in `python-libraries/nanover-openmm/src/nanover/openmm/serializer.py` we state that we can optionally save the OpenMM serialized state to our .xml files, but a closer inspection of the function revealed that this is currently untrue. But it would be excellent for it to be true!

hjstroud:
In OpenMM I think they achieve this through the `saveState` function, [see here](http://docs.openmm.org/7.0.0/api-python/generated/simtk.openmm.app.simulation.Simulation.html#simtk.openmm.app.simulation.Simulation.saveState)

hjstroud:
A further thought on this: If by saving the state we save the positions (and velocities, parameters, etc.), we ought to specify that when loaded the positions should be read from the state rather than the PDB.

PR #285, opened 2024-10-31
Save/load state in xml serializer

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/283#issuecomment-2450206980

### DIFF
--- a/python-libraries/nanover-omni/tests/openmm_simulation_utils.py
+++ b/python-libraries/nanover-omni/tests/openmm_simulation_utils.py
@@ -171,7 +171,7 @@ def serialized_simulation_path(basic_simulation, tmp_path):
     """
     Setup an XML serialized simulation as a temporary file.
     """
-    serialized_simulation = serializer.serialize_simulation(basic_simulation)
+    serialized_simulation = serializer.serialize_simulation(basic_simulation, save_state=True)
     xml_path = tmp_path / "system.xml"
     with open(str(xml_path), "w") as outfile:
         outfile.write(serialized_simulation)
@@ -183,7 +183,7 @@ def basic_simulation_xml(basic_simulation):
     """
     Generate an XML serialized simulation from the basic test simulation.
     """
-    xml_string = serializer.serialize_simulation(basic_simulation)
+    xml_string = serializer.serialize_simulation(basic_simulation, save_state=True)
     return xml_string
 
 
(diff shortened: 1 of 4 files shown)


Ragzouken:
```
<State openmmVersion="8.1.1" stepCount="0" time="0" type="State" version="1">




        <PeriodicBoxVectors>




                <A x="15" y="0" z="0"/>




                <B x="0" y="15" z="0"/>




                <C x="0" y="0" z="15"/>




        </PeriodicBoxVectors>




        <Parameters/>




        <Positions>




                <Position x="0" y="0" z="0"/>




                <Position x="5.288000106811523" y="1.6100000143051147" z="9.359000205993652"/>




                <Position x="2.0510001182556152" y="8.239999771118164" z="-6.785999774932861"/>




                <Position x="-10.6850004196167" y="-.5370000004768372" z="1.9210000038146973"/>




                <Position x="0" y="0" z="5"/>




                <Position x="5.288000106811523" y="1.6100000143051147" z="14.359000205993652"/>




                <Position x="2.0510001182556152" y="8.239999771118164" z="-1.7860000133514404"/>




                <Position x="-10.6850004196167" y="-.5370000004768372" z="6.921000003814697"/>




        </Positions>




        <Velocities>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




                <Velocity x="0" y="0" z="0"/>




        </Velocities>




        <IntegratorParameters/>




</State>

```

hjstroud:
A sensible test system to check that everything works as expected would be the single atom system using the velocity Verlet integrator with some force applied:

1. Run for _n_ steps with some defined force and record the velocity
2. Save the simulation with `save_state=True` to a file
3. Load the simulation from the saved file
4. Assert that the velocity in the loaded simulation equals that of the simulation before saving

It would be worth seeing if saving the state in this way also saves any forces being applied to the system...

hjstroud:
> A sensible test system to check that everything works as expected would be the single atom system using the velocity Verlet integrator with some force applied:
> 
> 1. Run for _n_ steps with some defined force and record the velocity
> 2. Save the simulation with `save_state=True` to a file
> 3. Load the simulation from the saved file
> 4. Assert that the velocity in the loaded simulation equals that of the simulation before saving
> 
> It would be worth seeing if saving the state in this way also saves any forces being applied to the system...

I have added a test that essentially implements this, but for a system of two methane molecules rather than a single atom system, includes multiple forces of different types instead of a single force, and uses a Langevin integrator rather than a Verlet integrator. This test does not test whether any forces applied to the system are also saved. 

I have also added the "save_state=True" flag to the existing tests for serializing/deserializing simulations, to assert that they still work correctly when this flag is set to True.

IMPORTANT NOTE: when writing this test, it was apparent that the velocities loaded from the serialized state were similar but not identical to those of the original state, hence the use of `pytest.approx`. However, it seemed that iterating the serializing/deserializing procedure on the loaded simulation _did_ produce identical values. @Ragzouken I don't think this is something that is related to what NanoVer is doing, it seems more likely to be an OpenMM thing, am I missing anything here?

Ragzouken (commented):
Tests look sensible!

Issue #290, opened 2024-11-05
Missing ability to start/stop recording a simulation within a session


rhoslynroebuck:
Currently we only offers the functionality to record an entire NanoVer session (from the point that the server is started to when it is closed). It would be great to add the ability to start and stop recordings within a single session.

**Generic use case:**
Start a server, player joins and puts the box in the correct place. Then start a script that loops over the following: (i) put simulation in desired position & start recording, (ii) wait for player to do their thing, (iii) stops recording.

**Proposed way of doing this**
@Ragzouken suggested that we do this by creating a class similar to the `InternalRunner` in the omni.py script (https://github.com/IRL2/nanover-protocol/blob/main/python-libraries/nanover-omni/src/nanover/omni/omni.py).

PR #291, opened 2024-11-05
Basic CLI for recording from server

Ragzouken:
Closes https://github.com/IRL2/nanover-protocol/issues/290


rhoslynroebuck:
I have tested this script and it is working for me when recording both with the python and rust servers. The only issue I'm having is during playback with the python server, where the `simulation_counter` increments by two each time a playback sim is loaded. I'm putting it here in case it's relevant, but I'm not sure whether this is a recording or playback issue. 

Notes: I am using `nanover-server` version #0.1.2763 (the latest version at time of writing).

Ragzouken:
A nicer interface: 
image

Ragzouken:
> I have tested this script and it is working for me when recording both with the python and rust servers. The only issue I'm having is during playback with the python server, where the `simulation_counter` increments by two each time a playback sim is loaded. I'm putting this issue here in case it's relevant, but I'm not sure whether this is a recording or playback issue.
> 
> Notes: I am using `nanover-server` version #0.1.2763 (the latest version at time of writing).

Let's consider this an issue with playback rather than recording (as you've said, playback works in the rust server, so the recording should be valid). I'll put up an issue

rhoslynroebuck:
@Ragzouken, just tested using `nanover-record` via the command line. 

Test protocol: start a NanoVer server, start the recording client with `nanover-record --path "test-recording`, run for a bit, then click Ctrl+C once to stop the recording, and a second time to stop the script.

I can start the script and stop the recording with Ctrl-C with no problems. However, when I press Ctrl+C again to stop the script, I get the below error and the recording files aren't created.

`(nanover-dev) ..\nanover-protocol> nanover-record --path "test-recording"
Connecting to localhost:38801.
Recording from server to test-recording/recording-1-20241216-162315.traj and test-recording/recording-1-20241216-162315.state. Press Ctrl-C to stop.
Done. Press Ctrl-C to quit or Return to begin recording again.Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "\miniforge3\envs\nanover-dev\Scripts\nanover-record.exe\__main__.py", line 7, in <module>
object address  : 000001C32744AF20
object refcount : 3
object type     : 00007FFAD720A3E0
object type name: EOFError
object repr     : EOFError()
lost sys.stderr`

The other thing is that we need documentation to go with this client. The documentation that comes up with `nanover-record --help` is great, but it would also be good to put an example usage on [this page](https://irl2.github.io/nanover-docs/concepts/recording.html) in the docs, with instructions on getting help in the command line. Also, would it be possible to add an example usage into the docstrings?  E.g. like at the top of this [recording script](https://github.com/IRL2/nanover-utils/blob/main/recording/nanover-client-recorder.py) in the `nanover-utils` directory.

Ragzouken:
Here's a PR for basic documentation: https://github.com/IRL2/nanover-docs/pull/161

rhoslynroebuck (approved):
Yeah this works for me, I've tested:
* Making single and multiple recordings for the same `nanover-record` client
* That these recordings work correctly during playback
* That you are able to specify the various arguments, i.e. `path`, `prefix`, `address`, and `autoconnect`.

This is an awesome feature, great work @Ragzouken!

PR #299, opened 2024-11-13
Include user forces and energies in the FrameData for ASESimulation

hjstroud:
This PR adds the total user energy to the frame, as well as the sparse user force array and corresponding sparse user index array (to label the indices of the atoms to which the forces apply). In implementing this, it seemed natural to move the function that creates the sparse arrays to a more general location (namely `imd_force.py`, where the iMD forces themselves are defined). This may not be the right location, and I welcome suggestions for more appropriate places!

For reference, I have tested this PR using a single-atom system, manually comparing the energies and forces with those from the OpenMM implementation (which give the same results). I have written a corresponding test for the single-atom system on the server to check the user energy equals the expected user energy, and also added a test that tests the sparse user forces written to the frame for a multi-atom system with multiple iMD forces applied.

This PR also separates the potential energy of the system and the potential energy associated with the user interaction (which were previously summed), so that they are delivered under the keys `energy.potential` and `energy.user.total`, respectively, as in our implementation of iMD for OpenMM.

### DIFF
--- a/python-libraries/nanover-omni/src/nanover/omni/ase.py
+++ b/python-libraries/nanover-omni/src/nanover/omni/ase.py
@@ -10,6 +10,7 @@ from ase.md.md import MolecularDynamics
 from nanover.app import NanoverImdApplication
 from nanover.ase.converter import (
     EV_TO_KJMOL,
+    ANG_TO_NM,
     ASE_TIME_UNIT_TO_PS,
     ase_atoms_to_frame_data,
 )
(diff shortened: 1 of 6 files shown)


python-libraries/nanover-omni/src/nanover/omni/ase.py
```diff
@@ -217,4 +219,13 @@ def make_regular_frame(self):
...
             frame_data.simulation_time = self.dynamics.get_time() * ASE_TIME_UNIT_TO_PS
 
+        # Add user forces and energies to frame
+        # (convert back from ASE internal units to NanoVer units)
+        user_sparse_indices, user_sparse_forces = get_sparse_forces(
+            self.atoms.calc.total_user_forces
+        )
+        frame_data.user_forces_sparse = user_sparse_forces * (EV_TO_KJMOL / ANG_TO_NM)
+        frame_data.user_forces_index = user_sparse_indices
+        frame_data.user_energy = self.atoms.calc.total_user_energy * EV_TO_KJMOL
```
Ragzouken, line 229: I think it probably makes sense to put this in https://github.com/IRL2/nanover-protocol/blob/c2a60435d495875b94a1b6677e7f3ee49051fb7b/python-libraries/nanover-ase/src/nanover/ase/converter.py#L65-L120 since it depends only on `atoms`

Ragzouken (approved):
Looks good, do as you like re: the comment I made.

python-libraries/nanover-omni/src/nanover/omni/ase.py
hjstroud: I think it was the fact that it assumed an iMD calculator that made me put it where it is. Maybe it would be sensible to add an if statement to check that the calculator is an iMD calculator, what do you think?

ludovicaisa (approved):
Going through the commits with @hjstroud, I approve these changes. Well done.

hjstroud:
> Note that with the new functionality, the behaviour when there are no user forces/energies is to send empty arrays/a value of zero to the client.

This still stands after the new changes!

PR #309, opened 2024-11-20
Attempt to fix developer install for pycharm IDE

Ragzouken:


hjstroud (approved):
Pretty sure this works for me on Mac! Nice one 👍

PR #310, opened 2024-11-20
Clear state on simulation switch in omni

Ragzouken:
Assumes that all avatars, play areas, interactions, and visualisations should be erased. This makes sense for demos in the general case at least.

- [x] Clear keys
- [x] Clear locks
- [x] Write tests


Ragzouken:
I've added a test and tested on standalone iMD VR that switching between a recording with interactions and a normal simulation doesn't leave avatars or interactions behind.

hjstroud (approved):
Looks great! We've been through this together in person, and discussed the possibility of in the future having certain keys that you may not wish to clear, hence only clearing the keys that are problematic at the moment. Nice one!

PR #314, opened 2024-11-22
Rework iMD in ASE

hjstroud:
This PR reworks the ASE implementation of iMD so that we can use it to perform quantitative iMD in the same way as we do in OpenMM currently. To do this, I have:

- Added an iMD force manager (`ImdForceManager`) class akin to that in the OpenMM iMD file, which handles calculation of the iMD forces and energies, and adds them to the frame data
- Reworked the `calculate` function in the `ImdCalculator` class so that it no longer uses the "_calculate_imd(...)" function but instead retrieves the iMD forces and energies from the `ImdForceManager`
- Adjusted how the potential energies and system forces are added to the frame data to use this new functionality
- Fixed the existing tests in `test_imd_calculator` and `test_imd_reset`: These are only initial fixes, and while the tests all pass I think it would be worth thinking about whether we need to alter them in future (once we've settled on exactly how this new form of quantitative iMD should look)

In addition to the existing tests for the `ImdCalculator`, I have tested this new implementation locally using the same single atom system that I used to test OpenMM (a single argon atom with a constant force applied), and can confirm that this works as expected in terms of:
- Energies (potential, kinetic and user)
- Forces (system and user)
- Positions
- Velocities 

I have not yet written corresponding tests for the ImdForceManager: @Ragzouken should these tests go in a new file or be included in the same file that the iMD calculator is tested?

### DIFF
--- a/python-libraries/nanover-ase/src/nanover/ase/converter.py
+++ b/python-libraries/nanover-ase/src/nanover/ase/converter.py
@@ -12,6 +12,7 @@ import itertools
 import numpy as np
 import numpy.typing as npt
 
+from nanover.ase.imd_calculator import ImdCalculator
 from nanover.trajectory import FrameData
 
 ANG_TO_NM = 0.1
@@ -118,9 +119,8 @@ def ase_to_frame_data(
         )
     if include_forces:
         data.particle_forces_system = ase_atoms.get_forces() * (EV_TO_KJMOL / ANG_TO_NM)
-        data.particle_forces_system -= ase_atoms.calc.get_property(
-            "interactive_forces", allow_calculation=False
-        ) * (EV_TO_KJMOL / ANG_TO_NM)
+        if isinstance(ase_atoms.calc, ImdCalculator):
+            data.particle_forces_system -= ase_atoms.calc.results["interactive_forces"] * (EV_TO_KJMOL / ANG_TO_NM)
 
     return data
 
@@ -267,10 +267,11 @@ def add_ase_state_to_frame_data(frame_data: FrameData, ase_atoms: Atoms):
     if energy is not None:
         frame_data.potential_energy = energy * EV_TO_KJMOL
         # Subtract iMD energy from total potential energy to obtain system potential energy
-        frame_data.potential_energy -= (
-            ase_atoms.calc.get_property("interactive_energy", allow_calculation=False)
-            * EV_TO_KJMOL
-        )
+        if isinstance(ase_atoms.calc, ImdCalculator):
+            frame_data.potential_energy -= (
+                    ase_atoms.calc.get_property("interactive_energy", allow_calculation=False)
+                    * EV_TO_KJMOL
+            )
     frame_data.kinetic_energy = ase_atoms.get_kinetic_energy() * EV_TO_KJMOL
 
 
--- a/python-libraries/nanover-ase/src/nanover/ase/converter.py
+++ b/python-libraries/nanover-ase/src/nanover/ase/converter.py
@@ -118,6 +118,9 @@ def ase_to_frame_data(
         )
     if include_forces:
         data.particle_forces_system = ase_atoms.get_forces() * (EV_TO_KJMOL / ANG_TO_NM)
+        data.particle_forces_system -= ase_atoms.calc.get_property(
+            "interactive_forces", allow_calculation=False
+        ) * (EV_TO_KJMOL / ANG_TO_NM)
 
     return data
 
@@ -263,6 +266,11 @@ def add_ase_state_to_frame_data(frame_data: FrameData, ase_atoms: Atoms):
         raise AttributeError("No calculator in atoms, so energy cannot be computed")
     if energy is not None:
         frame_data.potential_energy = energy * EV_TO_KJMOL
+        # Subtract iMD energy from total potential energy to obtain system potential energy
+        frame_data.potential_energy -= (
+            ase_atoms.calc.get_property("interactive_energy", allow_calculation=False)
+            * EV_TO_KJMOL
+        )
     frame_data.kinetic_energy = ase_atoms.get_kinetic_energy() * EV_TO_KJMOL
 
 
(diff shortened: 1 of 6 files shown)


python-libraries/nanover-ase/src/nanover/ase/converter.py
```diff
@@ -118,6 +119,10 @@ def ase_to_frame_data(
         )
     if include_forces:
         data.particle_forces_system = ase_atoms.get_forces() * (EV_TO_KJMOL / ANG_TO_NM)
+        if isinstance(ase_atoms.calc, ImdCalculator):
+            data.particle_forces_system -= ase_atoms.calc.results[
+                "interactive_forces"
+            ] * (EV_TO_KJMOL / ANG_TO_NM)
```
Ragzouken, line 125: Let's move this to its own function like the others for consistency.

python-libraries/nanover-ase/src/nanover/ase/converter.py
```diff
@@ -118,6 +119,10 @@ def ase_to_frame_data(
         )
```
Ragzouken, line 119: This one too while we're here.

python-libraries/nanover-omni/src/nanover/omni/ase.py
```diff
@@ -220,28 +231,8 @@ def make_regular_frame(self):
...
-                    self.atoms.calc.results["interactive_forces"]
-                ) * (EV_TO_KJMOL / ANG_TO_NM)
-        user_sparse_indices, user_sparse_forces = get_sparse_forces(
-            self.atoms.calc.results["interactive_forces"]
-        )
-        frame_data.user_forces_sparse = user_sparse_forces * (EV_TO_KJMOL / ANG_TO_NM)
-        frame_data.user_forces_index = user_sparse_indices
+        # to NanoVer units)
+        if self.imd_force_manager is not None:
+            self.imd_force_manager.add_to_frame_data(frame_data)
```
Ragzouken, line 236: Better to assert is not None at the beginning of make_regular_frame as it should always exist.

python-libraries/nanover-ase/src/nanover/ase/imd_calculator.py
```diff
@@ -10,14 +10,101 @@
...
+        # Get the current interactions from the iMD service (if any)
+        interactions = self.imd_state.active_interactions
+
+        # Determine whether interactions are the same as previous, and if
+        # not call _reset_velocities in ImdCalculator
+        if interactions != self._previous_interactions:
+            self.call_reset_velocities = True
+
+        # Not sure what the line below is here for? Comment out for now?
+        # self._reset_velocities(atoms, interactions, self._previous_interactions)
```
Ragzouken, line 67: Remember to delete.

python-libraries/nanover-ase/src/nanover/ase/imd_calculator.py
```diff
@@ -10,14 +10,101 @@
...
+        """
+        Get the forces to apply from the iMD service and communicate them
+        """
+        # Get the current interactions from the iMD service (if any)
+        interactions = self.imd_state.active_interactions
+
+        # Determine whether interactions are the same as previous, and if
+        # not call _reset_velocities in ImdCalculator
+        if interactions != self._previous_interactions:
+            self.call_reset_velocities = True
```
Ragzouken, line 64: I think ditch this check and field and assume it's always true -- reset velocities performs its own checks on if interactions need to be reset or not.

Ragzouken: On further discussion: set it to True regardless -- it's equivalent to signalling that forces have changed since the last reset.

python-libraries/nanover-ase/tests/test_imd_calculator.py
```diff
@@ -126,6 +133,8 @@ def test_one_interaction(
         with client_interaction(imd_client, interact_c):
             time.sleep(0.1)
             assert len(imd_calculator.interactions) == 1
+            # Update interactions in force manager
+            imd_calculator._imd_force_manager.update_interactions()
```
Ragzouken, line 137: This suggests that `update_interactions` should be available as a method on `imd_calculator` too -- without the caller having to be aware of the existence of the force manager.

Ragzouken: Perhaps also `add_to_frame_data` and then forget the `ImdForceManager` exists after construction and utilise it only via the `ImdCalculator` which it has been given to.

hjstroud:
After going through this with @Ragzouken earlier today, I have made some changes that basically make it so that the iMD interaction updates and addition of iMD data to the frame data occur via the `ImdCalculator` class directly, rather than via the `ImdForceManager`. I have refactored the tests to accommodate this change, and these tests should be sufficient to confirm that everything works as expected (at least for now).

Ragzouken:
> After going through this with @Ragzouken earlier today, I have made some changes that basically make it so that the iMD interaction updates and addition of iMD data to the frame data occur via the `ImdCalculator` class directly, rather than via the `ImdForceManager`. I have refactored the tests to accommodate this change, and these tests should be sufficient to confirm that everything works as expected (at least for now).

Latest changes look good!

python-libraries/nanover-ase/src/nanover/ase/imd_calculator.py
```diff
@@ -168,45 +248,47 @@ def calculate(
...
-        # convert back to ASE units (eV and Angstroms).
-        energy = energy_kjmol * ev_per_kjmol
-        forces = forces_kjmol * ev_per_kjmol / converter.NM_TO_ANG
-
-        # update previous interactions for next step.
-        self._previous_interactions = dict(interactions)
+    def update_interactions(self):
+        """
+        Update the iMD interaction energies and forces (in ASE units)
+        via the ImdForceManager.
```
Ragzouken, line 277: Should indicate this also applies velocity reset if applicable?

Ragzouken (approved):

PR #318, opened 2024-11-25
Calculate user work done in `ASESimulation`

hjstroud:
This PR adds the functionality to calculate the work done by the user on the molecular system to the ASESimulation class. It does so in the same way as implemented for OpenMM. As both `ASESimulation` and `OpenMMSimulation` classes need to calculate this quantity, the function `add_contribution_to_work` that was a function of the `OpenMMSimulation` class has now been moved to `imd_force.py`, and the calculation of work done in `OpenMMSimulation` has been refactored to reflect this.

So far, I have tested this PR locally using the single-atom system that I have been using to test previous PRs (consisting of a single Ar atom, to which a constant force is applied for a specified amount of time). This produces the same results as the OpenMM implementation, and the OpenMM implementation produces the same results as it did previously with the same system.

I have added corresponding tests to confirm that the ASE implementation works as expected. These tests take the same form as the existing tests that check the work done for OpenMM.

### DIFF
--- a/python-libraries/nanover-omni/src/nanover/omni/ase.py
+++ b/python-libraries/nanover-omni/src/nanover/omni/ase.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from typing import Optional, Any, Protocol
 
 import numpy as np
+import numpy.typing as npt
 from ase import Atoms
 from ase.calculators.calculator import Calculator
 from ase.md import MDLogger
@@ -12,6 +13,7 @@ from nanover.ase.converter import (
     EV_TO_KJMOL,
     ASE_TIME_UNIT_TO_PS,
     ase_atoms_to_frame_data,
+    ANG_TO_NM,
 )
 from nanover.ase.imd_calculator import ImdCalculator, ImdForceManager
 from nanover.ase.wall_constraint import VelocityWallConstraint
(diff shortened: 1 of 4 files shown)


python-libraries/nanover-omni/src/nanover/omni/ase.py
```diff
@@ -86,6 +88,11 @@ def __init__(self, name: Optional[str] = None):
         self.frame_index = 0
         self.ase_atoms_to_frame_data = ase_atoms_to_frame_data
 
+        self.work_done: float = 0.0
+        self._work_done_intermediate: float = 0.0
+        self.prev_imd_forces: Optional[np.ndarray] = None
+        self.prev_imd_indices: Optional[np.ndarray] = None
```
Ragzouken, line 94: Change these to underscore vars because they seem to be internal members.

python-libraries/nanover-omni/tests/test_ase.py
```diff
@@ -329,3 +329,48 @@ def test_imd_force_unit_conversion(example_ase_app_sim_constant_force_interactio
...
+
+    # For a simulation with a frame interval of 5 simulation steps and a simulation
+    # step size of 0.5 fs, using a constant force to accelerate the atom at 1 nm ps-1
+    # (for Ar this is a force of 40 kJ mol-1 nm-1), after 401 steps the work done on
+    # an Argon atom should be 20.0 kJ mol-1 analytically. Allowing for numerical error,
+    # the expected value should be slightly greater than 20.0 kJ mol-1
+
+    app, sim = example_ase_app_sim_constant_force_interaction
+
+    print(sim.atoms.get_masses())
```
Ragzouken, line 348: Remove debug print

Ragzouken (approved):
Some comments, otherwise looks good!

PR #323, opened 2024-11-28
Fix for OpenMM reset

Ragzouken:
Fixes https://github.com/IRL2/nanover-server-py/issues/322

Needs a test so we don't accidentally introduce this again. Also maybe I can find in the older code why this never happened before.

### DIFF
--- a/python-libraries/nanover-omni/src/nanover/omni/openmm.py
+++ b/python-libraries/nanover-omni/src/nanover/omni/openmm.py
@@ -110,8 +110,11 @@ class OpenMMSimulation:
         assert self.simulation is not None and self.checkpoint is not None
 
         self.app_server = app_server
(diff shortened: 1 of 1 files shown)


hjstroud (approved):
We have talked about this in person already, but this looks great! Well done for getting this fixed so fast, and for spotting what the problem was 🚀 Fine by me to include the HIV complex file for testing for now, we may wish to rethink this in the future.

ludovicaisa (approved):
I have checked this branch running PC-VR and Standalone setup with hiv_complex and other protein-ligand systems from this [directory](https://github.com/IRL2/nanover-server-py/tree/main/examples/basics/openmm_files).
I confirm that reset command works as expected from VR interface or jupyternotebook, the proteins structure is maintained, resolving cartoon rendering's [issue](https://github.com/IRL2/nanover-server-py/issues/192) .

PR #325, opened 2024-12-03
Simplify dcd_reporter closing in notebook

Ragzouken:
May improve https://github.com/IRL2/nanover-server-py/issues/233


ludovicaisa (commented):
I checked the notebook again and there are not missing frames, even chainging the number of timesteps.
What i suggest is to decrease the number of timesteps for storing frames, so that in the corresponding trajectory file you can see a more continuous simulation and not occuring in molecules "jumps" in the coordinates space: such as `dcd_reporter = app.DCDReporter('output.dcd', 50)` .

PR #326, opened 2024-12-03
changed nanotube file to the one from subtle game

Mohamed-Dhouioui:
This pull request includes updates to `examples/ase/openmm_files/nanotube.xml`,  `examples/openmm/openmm_files/nanotube.xml`, and `examples/basics/openmm_files/nanotube.xml` files which are now a copy of the one used in Subtle game.


hjstroud:
We discussed this PR in the dev call and decided that @rhoslynroebuck, @ludovicaisa and @hjstroud are going to handle updating these files, but need to talk about how we want to simulate each of these systems, where all of the parameters come from, etc. For this reason, I think it will be easier to create a new PR for this at the appropriate time and close the present PR.

PR #327, opened 2024-12-09
Attempt to fix simulation reset for openmm

Ragzouken:


Ragzouken:
This fix doesn't work for the intended case.

PR #330, opened 2024-12-09
Remove ASEOpenMMSimulation class

hjstroud:
This PR alters the following tutorial notebooks in the `examples` directory so that they use the `ASESimulation` class in place of the `ASEOpenMMSimulation` class when setting up ASE simulations using and OpenMM calculator:

- `ase_openmm_graphene.ipynb`
- `ase_openmm_nanotube.ipynb`

It also fixes the `ase_openmm_neuraminidase.ipynb` notebook, which was not running properly before.

I have tested these changes locally on my computer, and the notebooks seem to run normally. I have also tested that you can connect to the servers in VR and see what you expect to see using the standalone APK, and everything seems to behave as expected. I have not tested this with PC-VR (as I am unable to), so if someone wanted to check everything works fine for that use case they are welcome to! However I think the tests I have conducted myself suffice.

EDIT: This PR now also removes the `ASEOpenMMSimulation` class itself, including the tests associated with it.


hjstroud:
@Ragzouken is on updating the figure in `examples/fundamentals/images` that contains `ASEOpenMMSimulation` (which is the last reference to `ASEOpenMMSimulation` that isn't the definition or an internal test)

Ragzouken:
I have updated the diagram to be for ASE instead of ASEOpenMM. I have also removed the simulation type itself and its scant suit of tests.

Ragzouken (approved):
Notebook changes look good to me!

PR #335, opened 2024-12-19
Improve nanover-omni tests

Ragzouken:
Reduce unwieldy fixtures and prepare to remove non-omni openmm.

Need to replicate from non-omni openmm tests:

- [x] test_force_manager_masses
- [x] test_report_frame_forces
- [x] test_sparse_user_forces
- [x] test_apply_interactions
- [x] test_remove_interaction_partial
- [x] test_remove_interaction_complete
- [x] test_sparse_user_forces_elements
- [x] test_velocities_and_forces
- [x] test_velocities_and_forces_single_atom

### DIFF
--- a/python-libraries/nanover-omni/src/nanover/omni/cli.py
+++ b/python-libraries/nanover-omni/src/nanover/omni/cli.py
@@ -155,7 +155,7 @@ def main():
 
     with initialise_runner(arguments) as runner:
         if len(runner.simulations) > 0:
-            runner.next()
+            runner.load(0)
 
         if arguments.rich:
             try:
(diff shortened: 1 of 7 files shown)


hjstroud (approved):
This looks great, well done! We have been through this in person, and have made all of the changes that I think are necessary. Nice one! 😄

hjstroud:
One more thing! It might be worth making sure that we test the ethane system to make sure that a bunch of the stuff we test for works on multi-atom multi-interaction systems as well as our single atom system (with constant force interaction for indices 0, 1, 4 and harmonic interaction for indices 4, 5, or something similar)

PR #336, opened 2024-12-19
Switch target action in docs

Ragzouken:
If we want to use https://github.com/IRL2/nanover-docs/pull/168


hjstroud (approved):
Looks fine to me!

PR #338, opened 2025-01-08
Remove NanoverImdReporter from OpenMM

Ragzouken:
Somehow forgot to remove this in the earlier request.


hjstroud (approved):
Seems sensible to remove this if it's unused, well caught!

PR #340, opened 2025-01-09
Examples of how to filter keys from recording files

Ragzouken:
e.g filter out problematic keys for https://github.com/IRL2/nanover-server-py/issues/329


Ragzouken:
This already exists in an actually working state in: https://github.com/IRL2/nanover-utils/blob/main/parsing-recordings/remove_sim_counter_from_trajectory.py

PR #347, opened 2025-01-14
Cleanup ASE's ImdCalculator

Ragzouken:
In an ASE calculator, it seems that `self.atoms` is supposed to be a copy of the last calculated atoms, whereas were using it to store the actual atoms object used by the dynamics.

Remove support for `ImdCalculator` that isn't provided `atoms` and `calculator` to start with. There seems to be no usecase for this and it was undertested.


python-libraries/nanover-server/src/nanover/ase/imd_calculator.py
```diff
@@ -42,7 +42,11 @@ def update_interactions(self):
         """
         Update the iMD interaction energies and forces (in ASE units).
         """
+        prev_interactions = self._current_interactions
```
hjstroud, line 45: Why do we want to save and return the previous interactions?

python-libraries/nanover-server/src/nanover/ase/imd_calculator.py
```diff
@@ -160,7 +163,12 @@ def __init__(
         self.reset_scale = reset_scale
         self._custom_temperature = None
         self._initialise_velocity_reset()
-        self._pbc_implemented = False
+
+        self._atoms = atoms
```
hjstroud, line 167: So does this make a copy of the atoms object passed in, like in an ASE calculator?

python-libraries/nanover-server/src/nanover/ase/imd_calculator.py
```diff
@@ -160,7 +163,12 @@ def __init__(
...
+
+        # Check whether the periodic boundary conditions defined in the atoms object are implemented
+        # for the iMD calculator.
+        get_periodic_box_lengths(atoms)
```
hjstroud, line 171: Just to check, this does essentially the same thing that was happening previously but just more neatly and earlier?

python-libraries/nanover-server/src/nanover/ase/null_calculator.py
```diff
@@ -14,4 +14,6 @@ def __init__(self, **kwargs):
     def calculate(
         self, atoms=None, properties=("forces", "energy"), system_changes=all_changes
     ):
-        self.results = {"forces": np.zeros((len(atoms), 3)), "energies": 0.0}
+        super().calculate(atoms, properties, system_changes)
```
hjstroud, line 17: May be a silly question, but could you explain what the `super()` does here? I assume it behaves the same way as it does for the `__init__` function?

hjstroud (commented):
I think most of these changes make sense to me. My only real worry is I don't know what the reason for making things optional was in the first place, so I am unsure whether making them mandatory now is a good idea, but if you understand it better @Ragzouken then I trust your judgement. I have left a few comments (mostly because I don't fully understand what's happened/why), and once those are cleared up I'm happy to review again

python-libraries/nanover-server/src/nanover/ase/imd_calculator.py
Ragzouken: Yes exactly. The periodic check would happen once at the time of first calculation (because atoms had not been set yet), now it happens once at construction because atoms are available from the start.

python-libraries/nanover-server/src/nanover/ase/imd_calculator.py
Ragzouken: Velocity reset uses them to figure out which interactions were cancelled in the update. We could instead return just the set of cancelled interactions. The difference from before is just that they were saved as a member and accessed directly, but I think it's cleaner to at least just return them at the point of update since they're only used there anyway.

python-libraries/nanover-server/src/nanover/omni/ase.py
```diff
@@ -136,14 +136,11 @@ def reset(self, app_server: NanoverImdApplication):
...
         self.imd_calculator = ImdCalculator(
             self.app_server.imd,
-            ImdForceManager(self.app_server.imd, self.atoms),
-            self.initial_calc,
+            calculator=self.initial_calc,
+            atoms=self.atoms,
             dynamics=self.dynamics,
         )
 
-        # assign imd calculator to atoms
```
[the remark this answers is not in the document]
Ragzouken: I looked at how the ASE calculators work and part of the base behavior is to do this when atoms are passed, so because we pass atoms now it will happen by default.

python-libraries/nanover-server/src/nanover/ase/null_calculator.py
Ragzouken: In general `super()` is a shorthand to get the parent class of this class, so for `class NullCalculator(Calculator):` it's equivalent to calling `calculate` on `Calculator` before doing the rest. 

The docs for `Calculator` say:
> The subclass implementation should first call this implementation to set the atoms attribute and create any missing directories.

That said, we don't call it in `ImdCalculator` and if we do it breaks for reasons I don't understand (but definitely something to do with the way it updates `self.atoms`. I will add a comment to that effect to `ImdCalculator`.

python-libraries/nanover-server/src/nanover/ase/imd_calculator.py
hjstroud: Fair enough!

python-libraries/nanover-server/src/nanover/omni/ase.py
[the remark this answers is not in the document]
hjstroud: excellent, no problem then!

hjstroud (approved):
Looks good to me, well done!

PR #359, opened 2025-01-23
Replace unmaintained netifaces with psutil

Ragzouken:


Ragzouken:
Promising start but I need to debug this on a mac when Harry is back.

Ragzouken:
Need to test for real on all platforms, just in case

PR #361, opened 2025-01-24
Utility function to iterate a recording as full frame and state dicti…

Ragzouken:
Utility function for iterating over a recording as a series of complete snapshots of frames and state dictionaries (instead of just the sparse set of updates/removals).

### DIFF
--- a/python-libraries/nanover-server/src/nanover/recording/reading.py
+++ b/python-libraries/nanover-server/src/nanover/recording/reading.py
@@ -22,24 +22,27 @@ MAGIC_NUMBER = 6661355757386708963
 FrameEntry = Tuple[int, int, FrameData]
 
 
-def iter_full_view(*, traj: Optional[PathLike[str]] = None, state: Optional[PathLike[str]] = None):
+def iter_full_view(
+    *, traj: Optional[PathLike[str]] = None, state: Optional[PathLike[str]] = None
+):
+    """
+    Iterate one or both of trajectory and state recording files, yield a timestamp and copies of both the current
+    aggregate FrameData and the current aggregate state dictionary.
+    """
     full_frame = FrameData()
     full_state = StateDictionary()
 
     for time, frame, update in iter_recording_files(traj=traj, state=state):
-        frame_reset = frame is not None and frame.values["index"] == 0
-
-        if frame_reset:
-            full_frame = FrameData()
-
         if frame is not None:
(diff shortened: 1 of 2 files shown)


Mohamed-Dhouioui (approved):
This is working fine as far as I can see. I have noticed that "work_done" first appeared in the third frame of the recordings (I'm not sure if we want it that way). Additionally, when running the tests I got a warning about the docstring from the temperature calculation function (I think the warning refers to the invalid escape sequence `\m` which should be fixed by making the docstring a raw string), should I raise it as an issue: 
```
python-libraries\nanover-server\src\nanover\ase\thermo.py:9
  C:\Users\Dhoui\Documents\nanover-protocol\python-libraries\nanover-server\src\nanover\ase\thermo.py:9: SyntaxWarning: invalid escape sequence '\m'
    """
```

Ragzouken:
> This is working fine as far as I can see. I have noticed that "work_done" first appeared in the third frame of the recordings (I'm not sure if we want it that way). Additionally, when running the tests I got a warning about the docstring from the temperature calculation function (I think the warning refers to the invalid escape sequence `\m` which should be fixed by making the docstring a raw string), should I raise it as an issue?:
> 
> ```
> python-libraries\nanover-server\src\nanover\ase\thermo.py:9
>   C:\Users\Dhoui\Documents\nanover-protocol\python-libraries\nanover-server\src\nanover\ase\thermo.py:9: SyntaxWarning: invalid escape sequence '\m'
>     """
> ```

The escape sequence warning should be fixed now with https://github.com/IRL2/nanover-server-py/pull/364.

Ragzouken:
> This is working fine as far as I can see. I have noticed that "work_done" first appeared in the third frame of the recordings (I'm not sure if we want it that way). Additionally, when running the tests I got a warning about the docstring from the temperature calculation function (I think the warning refers to the invalid escape sequence `\m` which should be fixed by making the docstring a raw string), should I raise it as an issue?:
> 
> ```
> python-libraries\nanover-server\src\nanover\ase\thermo.py:9
>   C:\Users\Dhoui\Documents\nanover-protocol\python-libraries\nanover-server\src\nanover\ase\thermo.py:9: SyntaxWarning: invalid escape sequence '\m'
>     """
> ```

Issue related to `work_done` not present in first frame: https://github.com/IRL2/nanover-server-py/issues/365

PR #367, opened 2025-01-30
docs: rename examples folder to tutorials

Ragzouken:
Closes https://github.com/IRL2/nanover-server-py/issues/366


rhoslynroebuck (approved):
Looks good to me! Just a note that others may have to delete the old "examples" directory in their cloned repo because it is empty so git can't do anything with them.

PR #376, opened 2025-02-05
Reset work and iMD forces when reseting simulations

Ragzouken:
Closes https://github.com/IRL2/nanover-server-py/issues/373

Resets internal data for work done and imd forces when simulations are reset. Expands existing work tests to check that they still hold after multiple sequential runs after resets. Adds a fix to the ImdForceManager that clears out any residual forces in the OpenMM force between resets.


ludovicaisa (approved):
All the changes look good to me.

PR #378, opened 2025-02-10
fix: ignore ctrl+c during shutdown

Ragzouken:
fixes https://github.com/IRL2/nanover-server-py/issues/377 and https://github.com/IRL2/nanover-server-py/issues/379

Ignore keyboard interrupt while shutting down the server in the CLI. On windows this can interrupt threading which leaves the terminal in a hung state.


rhoslynroebuck (approved):
Yes this works for me!

PR #382, opened 2025-02-17
Add hyperballs to visualisations tutorial

Ragzouken:
Closes https://github.com/IRL2/nanover-server-py/issues/381

Image is low quality, ideally we replace it before merging this PR.


Mohamed-Dhouioui (commented):
This looks good to me. I took the liberty of updating the image with a white background for consistency with the other examples. Let me know if it is adequate or not.
One thing to note is that one test failed, and I belive it failed due to a 404 Not Found error when trying to download a specific version of the libssl1.1 package from the Ubuntu security repository. I think the required file is no longer available there.

Ragzouken:
Thanks. We'll ignore the test because it's either fixed on main or needs to be fixed on main if it fails there. Image looks good.

PR #391, opened 2025-04-03
Handle PBC in Openmm

ludovicaisa:
https://github.com/IRL2/nanover-server-py/issues/385


ludovicaisa:
In openmm.py was added a new property of the simulation which enforces the periodic boundary conditions if set to True.
We confirmed it works in VR but we need to create an automated test.

python-libraries/nanover-server/tests/test_omni/test_openmm.py
```diff
@@ -543,3 +545,38 @@ def test_velocities_and_forces_single_atom():
         assert frame.particle_velocities[i] == pytest.approx(
             expected_velocities, abs=1e-7
         )
+
+
+def test_pbc_enforcement():
+    """
+    Test that PBC wrapping of positions defaults to off, doesn't wrap positions when off, and does wrap positions when on.
+    """
+    omm_sim = build_basic_simulation()
```
ludovicaisa, line 554: I don't think that `build_basic_simulation` is the proper case to test this feature, as the charges of the particles are set to zero and the box size is large. I don't expect the methane to escape the box within a few simulation steps.

Ragzouken: It does actually resize the PBC later in the test, but I've moved the line here so that it's clearer that it's happening. I went with 1x1x1 which is sufficient for `build_basic_simulation` to have some positions that get wrapped depending on whether PBC is on or off.

To be sure that it has an effect, the test has a check both that with PBC off some positions are outside the box, and that with PBC on there are no longer particles outside the box, so we can be sure that the test system does actually have some particles outside the box.

python-libraries/nanover-server/src/nanover/omni/openmm.py
```diff
@@ -75,6 +75,8 @@ def __init__(self, name: Optional[str] = None):
         """Include particle forces in frames."""
         self.platform_name: Optional[str] = None
         """Name of OpenMM platform to use at the time the system is loaded from XML."""
+        self.enforce_pbc = False
```
Ragzouken, line 78: Maybe we can consider a different name for this. "enforce" sounds a little strong when it doesn't actually affect the simulation, just how the frames are reported--something more like `wrap_positions` would be more accurate? Or `wrap_molecules` or something.. it wraps whole objects based on COM but I don't know specifically how it groups things..

python-libraries/nanover-server/tests/test_omni/test_openmm.py
ludovicaisa: If this is the case then the testing is fine for me.

ludovicaisa (commented):
It looks good to me.

PR #401, opened 2025-04-15
Fix universes from recording function

Ragzouken:
I don't know why, but there can be strange resets in recordings where particle positions are present initially but topology information isn't. So we need to skip initial frames until both are present for this conversion.


rhoslynroebuck:
This fixes the problem I was having when processing SubtleGame data, thanks @Ragzouken !

PR #402, opened 2025-04-15
Split recordings

Ragzouken:


Ragzouken:
Sometimes the split .traj is empty (because things reset immediately?) but because there's preceding state we can't just discard it. Need nicer warning in recording_to_universe and maybe we can skip recordings without any meaningful traj?

Ragzouken:
The state splitting wasn't working at all but should now be fixed

rhoslynroebuck:
I tried to run the example, but run into an error.

command: `nanover-split-recording murkyBittern1.state murkyBittern1.traj`

error:
` Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "envs\nanover-dev\Scripts\nanover-split-recording.exe\__main__.py", line 7, in <module>
  File "nanover-server-py\python-libraries\nanover-server\src\nanover\recording\split_cli.py", line 209, in main
    if args.on_key_change is not None:
       ^^^^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'on_key_change'`

PR #406, opened 2025-04-24
Error handling for invalid interactions (particle indexes out of bounds)

Ragzouken:


Ragzouken:
@rhoslynroebuck here I have updated the iMD force code to throw a more descriptive exception when the index is out of bounds, and updated the omni runner so that if an exception occurs during the running of the simulation it will pause and set `system.simulation.exception` in the frame data (as well as printing on the terminal).

Issue #407, opened 2025-04-30
OpenMM implementation is enforcing PBCs in non-periodic simulations when serializing


hjstroud:
The expected behaviour of the OpenMM implementation is that the OpenMMSimulation class and the NanoVer OpenMM XML files saved using the serializer respect the periodicity of the simulation, i.e. match whether or not periodic boundary conditions are used in the simulation itself. However, recent changes have meant that the periodic box that OpenMM simulations contain as a default (but do not necessarily use as a default) is _always_ enforced when saving, effectively making non-periodic simulations periodic and leading to unexpected behaviour: I have not checked the behaviour for OpenMM simulations themselves. We need to change the functionality of the serializer (at least, possibly the OpenMMSimulation class too) to check the PBCs of the simulation when saving (and potentially loading for the OpenMMSimulation class), and use this to determine whether or not to use PBCs on the NanoVer side of things.

hjstroud:
As a side note, this issue wasn't caught by tests previously, so we also need to write tests to make sure the periodicity of the simulation is respected.

Ragzouken:
Do you know if this is the thing we should be using as a guide for whether to wrap coordinates or is there something else?

![Image](https://github.com/user-attachments/assets/a904bf20-5255-4302-a6e2-9c9821292a17)

hjstroud:
As far as I am aware, the part in the try statement is the correct thing. It's certainly the thing I use to check stuff in my own work!

PR #408, opened 2025-05-05
Save pbc

ludovicaisa:
Solving issue #407 by adding a new parameter to enforce PBC when saving a .xml simulation file

### DIFF
--- a/python-libraries/nanover-server/src/nanover/openmm/serializer.py
+++ b/python-libraries/nanover-server/src/nanover/openmm/serializer.py
@@ -50,7 +50,7 @@ from .imd import populate_imd_force
 ROOT_TAG = "OpenMMSimulation"
 
 
-def serialize_simulation(simulation: app.Simulation, save_state=False) -> str:
+def serialize_simulation(simulation: app.Simulation, save_state=False, pbc_wrapping=False) -> str:
     """
     Generate an XML string from a simulation.
 
@@ -65,7 +65,7 @@ def serialize_simulation(simulation: app.Simulation, save_state=False) -> str:
 
     # Extract the PDB
     positions = simulation.context.getState(
-        getPositions=True, enforcePeriodicBox=True
+        getPositions=True, enforcePeriodicBox=pbc_wrapping
     ).getPositions()
     pdb_content = StringIO()
     app.PDBxFile.writeFile(simulation.topology, positions, pdb_content)
@@ -93,7 +93,7 @@ def serialize_simulation(simulation: app.Simulation, save_state=False) -> str:
                 getVelocities=True,
                 getParameters=True,
                 getIntegratorParameters=True,
-                enforcePeriodicBox=True,
+                enforcePeriodicBox=pbc_wrapping,
             )
         )
         state_document = parseString(state_xml_str)
--- a/python-libraries/nanover-server/src/nanover/openmm/serializer.py
+++ b/python-libraries/nanover-server/src/nanover/openmm/serializer.py
@@ -50,12 +50,15 @@ from .imd import populate_imd_force
 ROOT_TAG = "OpenMMSimulation"
 
 
-def serialize_simulation(simulation: app.Simulation, save_state=False, pbc_wrapping=False) -> str:
+def serialize_simulation(
+    simulation: app.Simulation, save_state=False, pbc_wrapping=False
+) -> str:
     """
     Generate an XML string from a simulation.
 
     :param simulation: The simulation to serialize.
     :param save_state: Save the present state of the simulation too.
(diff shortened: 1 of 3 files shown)


Ragzouken:
Looks good, but you should add a test that verifies that it works as expected in both cases. This can go in the existing `test_serializer.py` file, and probably will be quite similar to `test_pbc_enforcement` in `test_openmm.py`.

Ragzouken:
To deal with the "Ruff to lint python code" test failing, run these commands to reformat the code: https://github.com/IRL2/nanover-server-py?tab=readme-ov-file#formatting--linting-tests

python-libraries/nanover-server/tests/test_openmm/test_serializer.py
```diff
@@ -124,3 +131,52 @@ def test_platform(basic_simulation_xml, platform):
...
+    The test deserializes two simulations with different setting of pbc and check if the positions are correct.
+    """
+    omm_sim = build_basic_simulation()
+    UNIT_SIMULATION_BOX_VECTORS = [[1, 0, 0], [0, 1, 0], [0, 0, 1]]
+    omm_sim.context.setPeriodicBoxVectors(*UNIT_SIMULATION_BOX_VECTORS)
+    sim = OpenMMSimulation.from_simulation(omm_sim)
+    sim.load()
+    with NanoverImdApplication.basic_server(port=0) as app_server:
+        sim.reset(app_server)
+        sim.advance_by_one_step()
```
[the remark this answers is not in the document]
Ragzouken: (but the two later in the test are necessary to get the coords out, so that's fine)

python-libraries/nanover-server/tests/test_openmm/test_serializer.py
```diff
@@ -124,3 +131,52 @@ def test_platform(basic_simulation_xml, platform):
...
+
+    with open("test-sim-pbc.xml", "w") as outfile:
+        outfile.write(serialize_simulation(omm_sim, pbc_wrapping=True))
+    with open("test-sim-no-pbc.xml", "w") as outfile:
+        outfile.write(serialize_simulation(omm_sim, pbc_wrapping=False))
```
Ragzouken, line 153: Instead of creating real files, you can do: 
```
wrapped_io = StringIO() # virtual file
wrapped_io.write(...) # write contents

wrapped_io.seek(0) # next read/write is from the start of the file
sim_pbc_obj = deserialize_simulation(wrapped_io.read()) # read from virtual file
```
This avoids polluting the filesystem, the time it takes to read/write files, and any accidents where the test doesn't write the file and passes because it read an old version or something.

hjstroud:
This is fixing the correct thing in the correct place, but I'm not sure it fixes it in the correct way. The periodicity of an OpenMM system is an intrinsic property of the simulation, and you can determine whether or not the system of the simulation is periodic via the `usesPeriodicBoundaryConditions()` function (see [here](http://docs.openmm.org/7.5.0/api-c++/generated/OpenMM.System.html#_CPPv4NK6OpenMM6System30usesPeriodicBoundaryConditionsEv)). Would it be more sensible to check whether the system itself is periodic or not at the time of saving using this function, and then save the system with or without PBCs accordingly?

Ragzouken (approved):
Test makes sense and seems to work. Looks good to me, let's discuss in the meeting tomorrow whether the default should remain `False` or if it should decide based on the system.

python-libraries/nanover-server/src/nanover/openmm/serializer.py
```diff
@@ -50,12 +50,15 @@
 ROOT_TAG = "OpenMMSimulation"
 
 
-def serialize_simulation(simulation: app.Simulation, save_state=False) -> str:
+def serialize_simulation(
+    simulation: app.Simulation, save_state=False, pbc_wrapping=False
```
Ragzouken, line 54: Let's default to `None` and in the case of `None` use the following code from OpenMM to decide which to do:
image

Ragzouken: Or not.. @hjstroud @ludovicaisa you two can decide which makes most sense for this

Ragzouken: Since you lose information enforcing PBC (diffusion), maybe defaulting to `False` is just fine.

hjstroud:
After our conversation and looking into the OpenMM source code, it seems that their `saveState()` function doesn't enforce PBCs (indeed it doesn't even have the option to do so), and thus the default option that @ludovicaisa has implemented matches this. I am happy to approve this PR, good work!

hjstroud (approved):

PR #411, opened 2025-06-24
Use `universe_from_recording` in recordings example

Ragzouken:


hjstroud (changes requested):
I have added some text and formatting to make the single universe vs multiple universe cases a bit more distinct. In addition, I have a couple of comments:

- I think it's great we have the functionality to parse both single and multiple trajectory recordings, and the way we do it now looks nice and clean, but I think in the changes we have lost some functionality. The old way of parsing the recordings included the ability to not convert NanoVer units to MDAnalysis units, which is an option that doesn't seem to be possible with the current implementation. I think this ought to be changed, either back to how we used to parse things or to bring this functionality into the current way of parsing things.
- I'm glad we have an example for the case of a multiple-universe recording, but I feel as though it would potentially be beneficial to demonstrate how to iterate through such a list of universes to extract information. For example, if someone tried to tie a knot in 17-alanine several times in the recording, it would be nice to iterate over the recordings too extract the energies/work done as functions of time and compare the results by plotting them. Maybe this is overkill though.

Ragzouken:
> I have added some text and formatting to make the single universe vs multiple universe cases a bit more distinct. In addition, I have a couple of comments:
> 
> * I think it's great we have the functionality to parse both single and multiple trajectory recordings, and the way we do it now looks nice and clean, but I think in the changes we have lost some functionality. The old way of parsing the recordings included the ability to not convert NanoVer units to MDAnalysis units, which is an option that doesn't seem to be possible with the current implementation. I think this ought to be changed, either back to how we used to parse things or to bring this functionality into the current way of parsing things.
> * I'm glad we have an example for the case of a multiple-universe recording, but I feel as though it would potentially be beneficial to demonstrate how to iterate through such a list of universes to extract information. For example, if someone tried to tie a knot in 17-alanine several times in the recording, it would be nice to iterate over the recordings too extract the energies/work done as functions of time and compare the results by plotting them. Maybe this is overkill though.

I've added the unit conversion option to the two functions now, I believe that's the only special option we providing, so it should cover everything. Providing a meaningful example of using multiple simulations is out of my wheelhouse, but I'm not against including one.

hjstroud:
Thanks for adding the unit conversion to these functions, that's a nice option to have. Is there already a test for the unit conversion, and if not can you add a test to make sure you get what's expected?

I have added information about how to retain NanoVer units when reading from MDAnalysis, and also added a line about how one can extract data from multiple recordings in the same way as for the single case.

In terms of a meaningful example, I just meant something like an alanine recording with a few resets, tying a knot in each one, just so we have something that we could parse and compare the results of. But it's non-essential, as long as it's clear how someone would extract the data from multiple recordings, so I am happy for the PR to go ahead without this.

PR #431, opened 2025-08-07
Example of user commands (simulation checkpointing)

Ragzouken:


hjstroud (approved):
This is fantastic! Really nice clean notebook, great to have a video demonstrating the functionality too! Nice one, approved 🙂

Issue #466, opened 2025-09-18
Incorrect behaviour of iMD force with PBCs in OpenMMSimulation


hjstroud:
At the moment, the onus is on the user to set the bool `use_pbc_wrapping` upon creation of an instance of the `OpenMMSimulation` class, even though the information about whether the underlying simulation uses periodic boundary conditions is available from the simulation itself. As a result, although the underlying MD simulation is still run using PBCs, when visualised without setting this bool to `True` the components of the molecular system can stray into other periodic images, and are not confined to the periodic box though clearly still interact as normal with the species in the original periodic box.

When `use_pbc_wrapping` is set to `True`, the molecules no longer stray from the periodic box (when they cross the boundary they appear on the other side of the box as expected). However, in this case the iMD force does not obey periodic boundary conditions (does not apply the force to the nearest periodic image of the molecule), such that if a user drags a molecule across a periodic boundary, the molecule appears on the other side of the cell as expected but the iMD force is applied to this new image rather than the closest periodic image. This leads to a discontinuity in the user force due to the apparent increase in the distance between the molecule and the user. 

The correct behaviour would be:
- The periodicity of the iMD simulation matches the OpenMM simulation (and is therefore read from the simulation directly, without user input)
- For periodic systems, the user force is applied to the closest periodic image of the molecule, regardless of where that molecule is rendered.

PR #467, opened 2025-09-18
Fix PBCs of iMD force in OpenMMSimulation

hjstroud:
This PR addresses Issue #466 by:
- Determining whether or not to apply PBCs to the iMD simulation based on the PBCs of the OpenMM simulation
- Applies the iMD force to the nearest periodic image of the atom with which the user is interacting when using PBCs, such that the iMD force obeys the same PBCs as the OpenMM simulation

It also adds a test to check that these PBCs are enforced correctly in an OpenMMSimulation constructed from an OpenMM simulation that uses PBCs.

I have tested the changes made on this branch by performing an iMD-VR simulation of a periodic OpenMM system, which yielded the correct behaviour expected of a periodic simulation based on the criteria outlined in Issue #466. I have also tested this on a non-periodic system, to make sure that none of these changes have affected the behaviour of non-periodic systems, and the system behaved normally (applying forces from over half a default periodic box vector away yielded forces in the same direction as the user force origin).


python-libraries/nanover-server/src/nanover/openmm/imd.py
```diff
@@ -38,14 +38,30 @@ def __init__(self, imd_state: ImdStateWrapper, imd_force: CustomExternalForce):
...
         # clear any residual forces in external force
         for particle in range(self.imd_force.getNumParticles()):
             self.imd_force.setParticleParameters(particle, particle, (0, 0, 0))
 
-    def update_interactions(self, simulation: Simulation, positions: np.ndarray):
+    def update_interactions(
+        self,
+        simulation: Simulation,
+        positions: np.ndarray,
+        pbc_vectors: Optional[np.ndarray] = None,
```
Ragzouken, line 51: We're switching to `T | None` instead of `Optional[T]` everywhere now, so better change this too

hjstroud: Nice, I was not fully aware of this! Should be fixed now.

python-libraries/nanover-server/src/nanover/openmm/imd.py
```diff
@@ -38,14 +38,30 @@ def __init__(self, imd_state: ImdStateWrapper, imd_force: CustomExternalForce):
...
             self._update_masses(simulation.system)
 
+        if self.periodic_box_lengths is None and pbc_vectors is not None:
+            # Check that the periodic cell vectors define an orthorhombic cell
+            assert np.all(pbc_vectors == np.diagflat(np.diag(pbc_vectors))), (
+                "The periodic box vectors do not correspond to an orthorhombic cell. "
+                "Periodic boundary conditions are currently only implemented "
+                "for orthorhombic systems."
+            )
+            self.periodic_box_lengths = np.diag(pbc_vectors)
```
Ragzouken, line 63: It seems that this function only cares about pbc vectors the first time you send it, and otherwise ignores it, so maybe it makes more sense to pass them into the constructor for `ImdForceManager` instead. It's constructed when the simulation resets (which happens initially too).

hjstroud: This makes sense to me. I've now refactored the code to pass the PBCs to the constructor for ImdForceManager so they're only set once. Have a look and see what you think!

PR #505, opened 2025-11-10
NGLView fixes

Ragzouken:


Ragzouken:
Most times two visuals are overlaid (one static, one correct)

PR #522, opened 2025-11-20
Adding support for iterating through MDAnalysis trajectories

TENeary:
Quick PR to add ability for MDAnalysisSimulation protocols to iterate through Universes.

### DIFF
--- a/python-libraries/nanover-server/src/nanover/app/cli/recorder_cli.py
+++ b/python-libraries/nanover-server/src/nanover/app/cli/recorder_cli.py
@@ -14,7 +14,7 @@ import time
 from nanover.essd import DiscoveryClient
 from nanover.utilities.cli import suppress_keyboard_interrupt_as_cancellation
 from nanover.websocket.client.app_client import get_websocket_address_from_hub
-from nanover.websocket.record import record_from_server
+from nanover.websocket.record import BackgroundRecordingContext
 
 
 def handle_user_arguments(args=None) -> argparse.Namespace:
@@ -48,7 +48,7 @@ def handle_user_arguments(args=None) -> argparse.Namespace:
         "--address",
         metavar="PROTOCOL://HOST:PORT",
         help="Connect to specific host address and port.",
-        default="ws://localhost:38801",
+        default="ws://127.0.0.1:38801",
     )
 
     parser.add_argument(
@@ -92,18 +92,25 @@ def main():
 
         print(f"Connecting to {address}.")
 
-        with suppress_keyboard_interrupt_as_cancellation() as cancellation:
-            executor, websocket = record_from_server(address, outfile)
+        interrupted = False
 
-            try:
+        with suppress_keyboard_interrupt_as_cancellation() as cancellation:
+            with BackgroundRecordingContext.from_address_to_path(
+                address=address, path=outfile
+            ) as recording:
                 print(f"Recording from server to {outfile}. Press Ctrl-C to stop.")
+                recording.future.add_done_callback(lambda _: cancellation.cancel())
                 cancellation.wait_cancellation(interval=0.01)
-            finally:
-                websocket.close()
-                executor.shutdown()
+                interrupted = not recording.future.done()
+
+        if not interrupted:
+            print("Recording ended by disconnection.")
+            break
 
         try:
-            input("Done. Press Ctrl-C to quit or Return to begin recording again.")
+            input(
+                "Recording ended by Ctrl-C. Press Ctrl-C to quit or Return to begin recording again."
+            )
         except KeyboardInterrupt:
             break
         except EOFError:
--- a/python-libraries/nanover-server/src/nanover/app/cli/server_cli.py
+++ b/python-libraries/nanover-server/src/nanover/app/cli/server_cli.py
@@ -19,6 +19,7 @@ from nanover.mdanalysis import UniverseSimulation
 from nanover.openmm import OpenMMSimulation
 from nanover.recording import PlaybackSimulation
 from nanover.utilities.cli import suppress_keyboard_interrupt_as_cancellation
+from nanover.websocket.client.app_client import NanoverImdClient
 from nanover.websocket.discovery import DiscoveryClient
 from nanover.websocket.record import record_from_runner
 
@@ -117,6 +118,14 @@ def handle_user_arguments(args=None) -> argparse.Namespace:
     )
 
     parser.add_argument(
+        "--remote",
+        nargs="?",
+        metavar="PROTOCOL://HOST:PORT",
+        help="Publish to remote server at specific host address and port.",
+        const="ws://localhost:38801",
+    )
+
+    parser.add_argument(
         "-n",
         "--name",
         help="Give a friendly name to the server.",
@@ -140,12 +149,23 @@ def get_all_paths(path_sets: Iterable[Iterable[str]]):
 
 @contextmanager
 def initialise_runner(arguments: argparse.Namespace):
-    with OmniRunner.with_basic_server(
-        name=arguments.name,
-        address=arguments.address,
-        port=arguments.port,
-        ssl=initialise_ssl(arguments),
-    ) as runner:
+    @contextmanager
+    def make_runner():
+        if arguments.remote is None:
+            with OmniRunner.with_basic_server(
+                name=arguments.name,
+                address=arguments.address,
+                port=arguments.port,
+                ssl=initialise_ssl(arguments),
+            ) as runner:
+                yield runner
+        else:
+            with NanoverImdClient.from_url(arguments.remote) as client:
+                print(f"Serving as client of remote server at {arguments.remote}")
+                with OmniRunner.from_client(client) as runner:
+                    yield runner
+
+    with make_runner() as runner:
         for path in get_all_paths(arguments.recording_entries):
             runner.add_simulation(PlaybackSimulation.from_path(path=path))
 
--- a/python-libraries/nanover-server/src/nanover/app/omni.py
+++ b/python-libraries/nanover-server/src/nanover/app/omni.py
@@ -100,7 +100,8 @@ class OmniRunner:
         """
         Print out basic runner info to the terminal.
         """
-        print(basic_info_string(self.app_server))
+        with suppress(Exception):
+            print(basic_info_string(self.app_server))
 
         list = "\n".join(
             f'{index}: "{simulation.name}"'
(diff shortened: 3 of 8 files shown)


TENeary:
Adding a comment to ensure there's a note regarding the compatibility between the Omni runner and this implementation of the Simulation and to make sure this is the intended functionality.
Currently the `play` signal from the Omni runner uses the `advance_by_seconds` func from the simulation protocol. 
The runner expects this to relate to the real world time elapsed between the previous sent frame but passing this as the change to simulation time doesn't make sense given simulation time/real world time.

Should Omni runner dt passed be altered or use a different method especially given all current protocols (this PR excluded) only advance to the next step regardless.

Ragzouken:
> Adding a comment to ensure there's a note regarding the compatibility between the Omni runner and this implementation of the Simulation and to make sure this is the intended functionality. Currently the `play` signal from the Omni runner uses the `advance_by_seconds` func from the simulation protocol. The runner expects this to relate to the real world time elapsed between the previous sent frame but passing this as the change to simulation time doesn't make sense given simulation time/real world time.
> 
> Should Omni runner dt passed be altered or use a different method especially given all current protocols (this PR excluded) only advance to the next step regardless.

As far as I remember, `advance_by_seconds` really only exists to make it possible to run the live recordings back in real time. Probably it would be better that each simulation just tracks that internally rather than expecting the server to tell it how time is passing--but exactly how/when frames are generated needs more thought it general.

For this PR it probably makes sense to just pick some sensible mapping factor of real world time to simulation time (possibly exposed as a property for convenience) and leave it at that. Recording/trajectory playback is underdesigned so I don't think there's much more to be done there in this PR.

Ragzouken:
I updated the `mdanalysis_trajectory` tutorial notebook to use this method, but playback is either static or too slow to see anything--picking a default time factor or some other change that makes the playback in that notebook proceed as expected would be good to demonstrate playback is working (using the VR step controls manually works fine).

TENeary:
> I updated the `mdanalysis_trajectory` tutorial notebook to use this method, but playback is either static or too slow to see anything--picking a default time factor or some other change that makes the playback in that notebook proceed as expected would be good to demonstrate playback is working (using the VR step controls manually works fine).

If you're using the 3TIC (ose) simulation in the notebook for testing, the lack of an observed change is due to my previous comment regarding advancing by seconds in real world vs. simulation time. Assuming the runner interval is set at roughly 30 FPS the simulation will attempt to jump 33 frames - as the simulation only has 24, it just loops back to the beginning so you don't see any change.

Ragzouken:
With your changes it works as expected in the notebook, which is good, but he logic of how it works seems quite convoluted and difficult to follow.

The two ways simulations are advanced are `advance_by_one_step` and `advance_by_seconds` -- one for stepping and one for unpaused playback. Both of these set `current_dt` which takes priority over `default_dt`. So then what is the purpose of the `time_to_step` property? 

`StepManager` has a `next_ts` method but it doesn't seem to be called by anything.

In `if (next_ts := next(self._time_to_step)) is None:` there is a lot going on, it would be better to break it up, but regardless is there ever a case where this is actually `None` anyway? 

I didn't realise this before but `iter(universe.trajectory)` just returns `universe.trajectory` so `_universe_iterator` can be removed and replaced with `universe.trajectory`. Then you can see that the `load` method should just be empty, which makes sense because the Universe is already ready at the time of creation.

`_next_frame` has an optional `reset` argument but it's never passed by anything, and even if `reset` were `False`, the next invocation of the method is going to restart the iterator either way (`next(universe.trajectory)` after the `StopIteration` exception just restarts the iterator).

Because this is quite difficult to reason about and there seem to be a few unused or misleading parts, I'm going to switch us to the simplified version I made in the other branch. Let's use that as a base and we can discuss if it's its missing some important behavior from the previous version.

TENeary:
> With your changes it works as expected in the notebook, which is good, but he logic of how it works seems quite convoluted and difficult to follow.
> 
> Because this is quite difficult to reason about and there seem to be a few unused or misleading parts, I'm going to switch us to the simplified version I made in the other branch. Let's use that as a base and we can discuss if it's its missing some important behavior from the previous version.

I disagree, the new implementation makes a number of odd choices regarding naming that do not conform to the existing Protocol the class is derived from and a number of regressive edits handling edge cases that will need to be addressed again.
I conceed that `StepTracker` may be clunky but removing it entirely without recapturing the original aim of its implementation seems excessive to me.

The previous implementation aimed to enable 3 distinct functionality via the `advance_by..` methods:
- `advance_by_seconds` maps real world time elapsing to sim time.
- `advance_by_one_step` step the simulation
- `advance_to_next_report` advance the simulation by user defined period of simulation time - controlled via StepManager (see notes below).

The new implenentation does not allow for this. I would argue playback mapping to real world seconds is a more niche usecase than simply advancing a fixed time step every frame. Further removing `advance_to_next_report` muddles the codebase in that every other `Simulation` type implements this method - rather its the only method they actually implement - so it not being consistent across all types is inconsistent. 
I also don't understand the rationale for renaming the "advance" methods to "seek" in general. While I agree this is better reflects their actual function it goes against the current naming scheme from the existing protocols.
This also makes any sort of generic interface with using these classes significantly more difficult to implement. 
If anything, it should be the implementation in this Simulation that sets the precedence for the first two methods as ASE and OpenMM are always ultimately calling the same `advance_to_next_report`.

> The two ways simulations are advanced are `advance_by_one_step` and `advance_by_seconds` -- one for stepping and one for unpaused playback. Both of these set `current_dt` which takes priority over `default_dt`. So then what is the purpose of the `time_to_step` property?
> 
The goal of the `StepManager` was to enable a method for user to define a default step size when using the `advance_to_next_report` - i.e. the `default_dt`. This parameter is accessed whenever `current_dt` is None. 
This allows the `time_to_step` for a default step time and optionally arbitrary steps to be used simultaneously. While I am open to an alternative implementation, I do think there is use cases for this.
Refer to above for why I believe that renaming methods is a poor idea without it being systematic.

> In `if (next_ts := next(self._time_to_step)) is None:` there is a lot going on, it would be better to break it up, but regardless is there ever a case where this is actually `None` anyway?
> 
This can be None when the default time to step (`default_dt`) and `current_dt` are  both None, giving a fallback of just advancing a single frame. The `next_ts` property is a means to grab the next intended time jump without exhausting the currently set `current_dt` which is converted to None after `next()` is called. While not called in the code it is useful for checking the intended time jump before it is done.

> I didn't realise this before but `iter(universe.trajectory)` just returns `universe.trajectory` so `_universe_iterator` can be removed and replaced with `universe.trajectory`. Then you can see that the `load` method should just be empty, which makes sense because the Universe is already ready at the time of creation.
> 
Happy to change `iter(...)` to just `universe.trajectory`. the argument against removing the `iter` statements is to prevent UB resulting frmo when a user still has a reference to the universe. They may move the trajectory frame and keeping helps to ensure the simulation is always set to the first timestep.
I don't see a reason to do a bunch of additional `next_frame`/`frame_count` calculations when all of this is already stored/managed by the simulation. To prevent UB wouldn't it make more sense to use MDA's internal data? Additionally, the only instance where it is actually necessary to track any time progressing is for the real world relating to fracitonal increases in simulation dt - i.e. essentailly advancing by a fraction of a frame.

> `_next_frame` has an optional `reset` argument but it's never passed by anything, and even if `reset` were `False`, the next invocation of the method is going to restart the iterator either way (`next(universe.trajectory)` after the `StopIteration` exception just restarts the iterator).
> 
The goal of `reset` in `_next_frame` is allow for cases where a user wants to halt a simulation playback at the end of its timecourse instead of running in perpetuity. Perhaps this functionality is better handled by a member variable so it can more easily set by the user given the current restrictions in the function signature of the Simulation Protocol. But I believe this functionality is important to retain.

Also, from what I understand of this current implementation, the `seek_to_time` also has (what I would consider) poor behaviour. If a user provides a time > simulation time the Simulation moves to some frame in the middle of the simulation. e.g. passing 130s into simulation with a total time of 100s would jump to the frame representing 30s.
With maybe the exception of mapping playback to seconds elapsed in the real world this doesn't really seem to be a sensible resolution. I would argue this should either return the last frame (or first, if looping) or raise some error.

TENeary:
For the minor stuff, e.g. frame jump computations - I don't have any super critical issues its mostly semantic issues or me attempting to future proof the code for future issues I see arising. 
I expect that I didn't make my intentions particularly clear in the previous comment.
Regarding these small issues:
- I expect I wasn't clear enough - I have no issues dropping the `universe_iterator` member - especially as all functionality can be handled by manipulating the underlying `.trajectory`.
- `seek_to_next_frame` is overengineered, in my opinion. It can simply be resolved with `next(universe.trajectory)` + some minor work to handle reaching the end of the trajectory. There is no need to compute frame indexes etc.
- The `iter` point I referred to is for the `load` method specifically. I intended to mean that the simulation should always be reset on load - i.e. it sets the expectation that anytime a simulation is loaded it will always start from the beginning.
- `...totaltime` bug is a minor off by 1 error resolved with `...totaltime + universe.trajectory.dt`
- I wasn't aware of the `reset` bug in `_next_frame` but it can simply be resolved by jumping to the last timestep after a `StopIteration` (e.g. `universe.trajectory[-1]`).

That said, I still believe there are 2 key issues to be resolved:

Regarding the 3 seperate functionalities: 
I strongly disagree that the advance by a set number of simulation time (`advance_to_next_report`) in my previous comment is unneeded. I would expect `advance_by_seconds` in a class handling MDA simulations to advance the simulation to the frame corresponding to the the time `dt` seconds ahead. I, and I expect for most other people, would see this as the intuitive option.
From the function name, you would expect the simulation to advance by some number of (nano)seconds - not some arbitrary multiple of this - and without reading some documentation its impossible to determine this is what is happening.
MDA simulations have no relation to real world time - the only reason this is an issue is because of the afformetioned `PlaybackSimulation`. 
This is an outlier in the simulation types as it is the only one that has any realation to the real world time elapsing - so I don't see why every other simulation type (that actually implements the method) should bend over backwards to accomodate it.
I would suggest that if this is absolutely required, the onus is on the `Runner` not the simulation to handle any discrepencies. Or, as I had previously implemented, allow for all 3 options (I'm open to alternative name suggestions).
Further, just because no one else has requested this functionality, doesn't mean its not useful - rather, as I had implemented it, I am implicitly asking to be able to do this.

Regarding handling jumping to a timepoint that is out of bounds for the simulation (e.g. asking to `seek_to_time(15)` when the simulation is only 10 ns long):
Again, I would argue that the current functionality makes the least sense and is only a product of having the `PlaybackSimulation` as the previous standard.
I have no issues with this being an option but in most contexts I don't believe this is the intuitive behaviour. I would suggest that jumping to the last timestep (or first) should at least be default (or maybe throw an OutOfBounds style error?).
Consider someone working with the trajectory - without some indicator of time also present (i.e. in VR) - its impossible to tell whether a change you are seeing responds to an actual step of said (nano)seconds. If it simply jumps to a point midway through the simulation (e.g. 5ns in this example) implying the simulation is longer than it is.
Rather, this behaviour should be controlled with some member/setter func to modify this as required.

PR #542, opened 2025-12-22
can package use uv?

Ragzouken:


Ragzouken:
No. Conda wants to install dependencies its way. Would have to switch to pixi or something, but it's a totally different workflow.

PR #547, opened 2026-02-19
Add dynamic bonding and molecule detection for ASESimulation class

hjstroud:
This PR adds two new features to the ASESimulation class, which are specifically intended for use in simulations in which the topology of the molecular system can change, e.g. when using a quantum chemical simulation package as the backend of the simulation through ASE. The features added by this PR are:

- Molecule detection: This uses the bonding topology to determine which molecular (i.e. bonded) species are present in the simulation, and assigns each bonded group to an individual residue and chain, which are named according to the elements they contain (at the moment, though this is not terribly elegant, and perhaps something like SMILES strings would be better)
- Dynamic topology updates: This calls a function to calculate the bonds in the system, compares these current bonds to the bonds in the previous frame, and if they differ causes a new topology frame to be sent (which includes detecting the new molecular species and updating the residues and chains in the system). This feature is optional, and can be toggled on and off during a simulation.

TESTING:


Ragzouken:
If possible it'd be good to include a notebook with a system where this dynamic bonding is demonstrated

